Issue #56, opened 2014-01-09
Just an idea: Why not make `LiftN` the requirement for applicatives rather than `ap`?


russellmcc:
It seems like in a language like JavaScript where currying isn't the norm (and might not be performant because of a lack of compiler help), it might be more useful to make something more like `LiftN` the requirement for applicative rather than `ap`.  It could even be backwards compatible for non-library code by just allowing `ap` to take multiple arguments like this:

```
MyApp.of(f).ap(a, b, c, d, e)
```

This would be fairly easy to implement for most applicatives I can think of, and would certainly be more convenient for the user and probably more performant, too.

If this doesn't match the aesthetics of fantasy-land (and my guess is it doesn't), maybe there should be a way for implementors to optionally provide a `liftN` method (under some friendlier name, maybe just `lift`) for performance reasons?

bergus:
I think we should revisit this (I want to implement it :-)).
Making a [`liftA2`](http://hackage.haskell.org/packages/archive/base/latest/doc/html/Control-Applicative.html#v:liftA2) (or [`liftF2`](http://hackage.haskell.org/package/semigroupoids-5.0.0.2/docs/Data-Functor-Apply.html#v:liftF2)) function the alternative definition of `Apply` seems very attractive (would be placed at `seq` in @phadej's diagrams). I guess that it's more often needed than `ap`. Of course they can be derived from each other, but making `lift` is a bit ugly in JS as it requires currying.
```js
function ap(af, ax) {
    function call(f, x) { return f(x); }
    return lift(call, af, ax);
}
function lift(f, ax, ay) {
    var f_curried = function(x) { return function(y) { return f(x, y); }; };
    return ax.map(f_curried).ap(ay);
}
```
I don't think we should make this a method though but rather a static function, and I cannot really envision a name for it either: `ax.apWith(f, ay)` sounds wrong because we're not applying `ax` but rather `f`, and I really don't want to build upon [`thrush`](http://hackage.haskell.org/package/data-aviary-0.4.0/docs/Data-Aviary-BirdsInter.html#v:thrush) or *T-combinator* (see [this SE question](http://programmers.stackexchange.com/q/198575/66652) for references).

----
[...]

Issue #24, opened 2013-04-14
Semver the spec


jden:
I'm starting to see a bunch of projects starting up implementing some version or other of this spec. Since it's changing quite a bit at this early stage, I propose adding a [semver](http://semver.org/) version number to allow for clarity of which version of the spec is being implemented.

PR #66, opened 2014-04-15
Make package export fantasy-fill

phadej:
Related to https://github.com/fantasyland/fantasy-land/issues/24 https://github.com/fantasyland/fantasy-land/issues/56

With the fantasy fill you can fill the gaps in your pure functional, monadic fantasies.

```js
var Id = require('./id.js');
var fantasyFill = require('./index.js');

function MyId(x) {
	this.value = x;
}

MyId.of = function (value) {
	return new MyId(value);
};

MyId.prototype.chain = function (action) {
	return action(this.value);
};

console.log('Id: ', fantasyFill(Id).join(", "));
console.log('MyId: ', fantasyFill(MyId).join(", "));

console.log('MyId 3 `map` \\x -> x * x: ', MyId.of(3).map(function (x) { return x * x; }));
```

```
Id:  Applicative, Apply, Chain, Functor, Monad, Monoid, Semigroup
MyId:  Applicative, Apply, Chain, Functor, Monad
MyId 3 `map` \x -> x * x:  { value: 9 }
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,53 @@
+function fantasyFill(constructor) {
+    "use strict";
+
+    var hasConcat = typeof constructor.prototype.concat === 'function';
+    var hasEmpty = constructor.empty || constructor.prototype.empty;
+
+    var hasOf    = typeof constructor.of === 'function';
+    var hasChain = typeof constructor.prototype.chain === 'function';
+    var hasAp    = typeof constructor.prototype.ap === 'function';
+    var hasMap   = typeof constructor.prototype.map === 'function';
+
+    // of + chain -> ap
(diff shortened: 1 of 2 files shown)


SimonRichardson:
@phadej now i've added the laws, would it be better that the laws are tested rather than just checking for a function?

```js
var hasChain = typeof constructor.prototype.chain === 'function' && laws.chain. associativity(constructor)(equality)(1);
```

Then we know for certain that if it quacks like a duck and sounds like a duck then it is a duck.

SimonRichardson:
It doesn't seem like this is getting any traction, if you want to keep it open, let me know.

PR #82, opened 2015-01-21
change Monoid#empty from nullary function to identity element itself

michaelficarra:
:sweat_smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,18 +86,18 @@ A value which has a Semigroup must provide a `concat` method. The
 A value that implements the Monoid specification must also implement
 the Semigroup specification.
 
-1. `m.concat(m.empty())` is equivalent to `m` (right identity)
-2. `m.empty().concat(m)` is equivalent to `m` (left identity)
+1. `m.concat(m.empty)` is equivalent to `m` (right identity)
+2. `m.empty.concat(m)` is equivalent to `m` (left identity)
 
-#### `empty` method
+#### `empty` value
 
-A value which has a Monoid must provide an `empty` method on itself or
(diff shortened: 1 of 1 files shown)


puffnfresh:
:neutral_face:

5outh:
I think I'm in favor of this, does it screw with anything? @puffnfresh

puffnfresh:
No preference. Can construct an argument either way.

SimonRichardson:
Question: should changes like this PR also fix all the other fantasy-land repos?

5outh:
On second thought, changing the spec from how it was may cause other peoples' projects to no longer conform to the spec and if there's no preference either way, I think we should keep it the way it is.

SimonRichardson:
@michaelficarra what are the reasons behind this i.e. what's the motivation other than it looks nicer?

joneshf:
We're semver'ing this thing, so we could just bump the version.

5outh:
I think @SimonRichardson has a point, we should keep all of the fantasy-repositories up to date with the latest version, which is kind of unfortunate.

Should we take this to a vote or something?

davidchambers:
This approach is incompatible with the current [Id implementation][1]:

> ```javascript
> // Monoid (value must also be a Monoid)
> Id.prototype.empty = function() {
>     return new Id(this.value.empty ? this.value.empty() : this.value.constructor.empty());
> };
> ```


[1]: https://github.com/fantasyland/fantasy-land/blob/master/id.js

SimonRichardson:
Can we close this, it doesn't seem to over any value and consider it could break implementations I don't see the point. I hope you don't mind?

Issue #83, opened 2015-02-05
Add Then (>>) to monad spec


TrevorBasinger:
For example:
```
Id.prototype.then = function (m) {
    return m;
}
```

```
Promise.prototype.then = function (m) {
    var promise = this;
    return new Promise (function (resolve) {
        return promise.fork (function (dontCare) {
            return m.fork (resolve);
        });
    });
}
```

Then you could have code like this:
```
somethingToDo (x) // Returns Future
    .chain (thisAlsoReturnsAFuture)
    .then  (justDoFinalFuture)
```

instead of:
```
somethingToDo (x) // Returns Future
    .chain (thisAlsoReturnsAFuture)
    .chain (konst (justDoFinalFuture))
```
Thoughts?

5outh:
If `then` is the same as `>>`, then:

```js
m.then(k) === m.chain(function () { return k; });
```

It may be worth mentioning that this is the case and that `then` can be useful, but I don't know if `then` should need to be implemented outright or appear as a requirement for any of the structures.

If you can think of a way to incorporate `then` into fantasy-land, go ahead and submit a PR! I'd love to see how it comes out on paper and would be happy to see mention of `>>` in the `chain` specification.

bergus:
`then` has a very specific meaning in promise libraries. Please avoid name clashes with those.

PR #84, opened 2015-02-05
Added 'then' (>>) specs

TrevorBasinger:
This is for issue #83 .

### DIFF
--- a/id.js
+++ b/id.js
@@ -32,10 +32,15 @@ Id.prototype.chain = function(f) {
     return f(this.value);
 };
 
+// Then
+Id.prototype.then = function (m) {
+  return this.chain (function (x) { return m; });
+};
+
 // Extend
 Id.prototype.extend = function(f) {
     return new Id(f(this));
-}
+};
 
 // Monad
 Id.of = function(a) {
@@ -45,6 +50,6 @@ Id.of = function(a) {
 // Comonad
 Id.prototype.from = function() {
     return this.value;
-}
+};
 
 if (typeof module == 'object') module.exports = Id;


michaelficarra:
This is not the spec, it is just a sample implementation. And I agree with @5outh from #83: this function should be derived, not implemented individually by each Monad.

5outh:
Right, the point of fantasy-land is to guide implementations of various structures. What I meant in #83 was that mentioning `then` in the specification as a derivable function is not necessarily a bad idea, as long as it goes along with the document nicely.

TrevorBasinger:
@5outh Oh, I think I understand now. @michaelficarra You're correct, I completely over looked the spec.

If you guys wouldn't mind, would you please expand on why you'd rather see it derived instead of implemented on monads? I'm still fairly new to FP. I'm curious why Haskell implements return and two forms of bind, but in this situation it doesn't make sense. (or is >> derived in Haskell somewhere?)

From a "using fantasyland" perspective, it seems like it would be something that is useful. Admittedly, Promises are the only use case I can think of at the moment. But I attribute that to my lack of experience.

michaelficarra:
It is derived in Haskell. Here is an excerpt from [the source](http://hackage.haskell.org/package/base-4.7.0.2/docs/src/GHC-Base.html#%3E%3E):

```hs
    {-# INLINE (>>) #-}
    m >> k      = m >>= \_ -> k
```

We're not saying it's not useful, but that it should only be implemented once, ever.

puffnfresh:
```js
function then(a, b) {
  return a.chain(function(ignored) {
    return b;
  });
}
```

This works for all structures which support Bind. This is the _fundamental_ point of monads - being able to write these functions only once. Code reuse!

PR #91, opened 2015-03-30
Clearer variable names in identity fn

sdwebster:
It's visually confusing for Applicative named 'a' to call an identity function that also has a variable named 'a'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ need to implement:
 
 * Functor's `map`; derivable as `function(f) { return this.of(f).ap(this); })}`
 
-1. `a.of(function(a) { return a; }).ap(v)` is equivalent to `v` (identity)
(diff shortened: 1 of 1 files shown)


davidchambers:
#90 didn't go anywhere, so you may meet resistance here. Good luck to you, @sdwebster!

sdwebster:
Hi David! Sorry I hadn't seen #90 when I made #91. I have no strong opinion on the overall naming scheme. #91 is simply a replacement in a single statement where there's variable confusion.

davidchambers:
I agree that the choice of names is confusing, but @puffnfresh may argue that names don't matter. ;)

puffnfresh:
The name doesn't matter but shadowing has benefits and drawbacks.

PR #98, opened 2015-07-13
add most.js

chicoxyzzy:


SimonRichardson:
> They also implement Fantasy Land Monoid, Functor, Applicative, and Monad.

Not to nit pick, but do you really implement `Applicative`. I'm not seeing a `of` function, can you point me to this?

PR #110, opened 2015-09-07
Use arrow functions in code snippets

### LINKED ISSUE #109: Use arrow functions
They are now [supported by latest Chrome, Opera, Edge, and Firefox](https://kangax.github.io/compat-table/es6/#arrow_functions). And I think most of JavaScript developers already familiar with them. Perhaps it makes sense to rewrite code snippets in spec using them.

A small example:

 1. `u.map(function(a) { return a; })` is equivalent to `u` (identity)
 2. `u.map(function(x) { return f(g(x)); })` is equivalent to `u.map(g).map(f)` (composition)

With arrow functions:

 1. `u.map(a => a)` is equivalent to `u` (identity)
 2. `u.map(x => f(g(x)))` is equivalent to `u.map(g).map(f)` (composition)

Should I make a PR?

rpominov:
Fixes #109

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,8 @@ its `constructor` object. The `empty` method takes no arguments:
 
 ### Functor
 
-1. `u.map(function(a) { return a; })` is equivalent to `u` (identity)
-2. `u.map(function(x) { return f(g(x)); })` is equivalent to `u.map(g).map(f)` (composition)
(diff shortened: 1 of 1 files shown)


CrossEye:
:+1:

5outh:
Looks good to me, it's more readable this way for sure (and looks much more like the Haskell counterparts)

PR #112, opened 2015-09-29
Fixed #111 - Corrected inconsistencies in id.js

### LINKED ISSUE #111: Example id.js is incorrectly commented
The spec in the `readme.md` currently defines these data structures with methods:
```
Apply - ap()
Applicative - of()
Monad
Traversable - sequence()
```

But when I look at the comments in the `id.js` file, the comments link some methods to different algebras:
```
Applicative - ap()
Monad - of()
Traversable - traverse()
```

As a beginner to algebraic data types this is very confusing to me, what is what? Maybe the `Id` container is out of date? For example, I'm currently wondering what the difference is between an Apply and an Applicative. If I implement `Container.prototype.ap`, is an instance of that container called an apply or an applicative? And would a Haskell programming give the same answer? Should it not be that obvious for senior functional programmers that this mistake could not be made? Or is this completely new ground in JS?

The [Identity container from ramda-fantasy](https://github.com/ramda/ramda-fantasy/blob/master/src/Identity.js) is incomplete as far as I can see, but nicely annotated and seems to be better in sync with the spec, am I correct?

I propose the spec and the example Identity type be aligned to avoid this kind of confusion. I think it's very important for them to stay in sync.

As an aside, I think it's a good idea to give the Id container a more prominent spot, like in the first few paragraphs, it's of great help next to a very abstract specification.

branneman:
The comments in the Identity container are now in sync with the spec, making for less confusion.

See #111

### DIFF
--- a/id.js
+++ b/id.js
@@ -27,13 +27,13 @@ Id.prototype.map = function(f) {
     return new Id(f(this.value));
 };
 
-// Applicative
+// Apply
 Id.prototype.ap = function(b) {
     return new Id(this.value(b.value));
 };
 
 // Traversable
-Id.prototype.traverse = function(f, of) {
+Id.prototype.sequence = function(f, of) {
     return f(this.value).map(function(y){ return new Id(y); });
 };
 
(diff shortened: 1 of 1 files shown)


5outh:
Thanks :+1:

CrossEye:
:+1:

Issue #113, opened 2015-10-24
sequence: disagreement between documentation and implementation


davidchambers:
Documentation:

> A value which has a Traversable must provide a `sequence` method. The `sequence` method takes one argument:
>
>     u.sequence(of)

Example implementation:

> ```javascript
> // Traversable
> Id.prototype.sequence = function(f, of) {
>     return f(this.value).map(function(y){ return new Id(y); });
> };
> ```

Does the method expect one argument or two?

/cc @DrBoolean

DrBoolean:
Looks like the implementation was left as traverse, but renamed to sequence. Should only take 1 arg

PR #114, opened 2015-10-24
not traverse, but sequence silly

DrBoolean:
#113 pointed out that I'd left this as the previous traverse implementation. This updates the example

### DIFF
--- a/id.js
+++ b/id.js
@@ -34,7 +34,8 @@ Id.prototype.ap = function(b) {
 
 // Traversable
 Id.prototype.sequence = function(of) {
-    return this.value.map(function(y){ return new Id(y); });
+    // the of argument is only provided for types where map might fail.
+    return this.value.map(Id.of);
 };
 
 // Chain
--- a/id.js
+++ b/id.js
@@ -33,8 +33,8 @@ Id.prototype.ap = function(b) {
 };
 
 // Traversable
-Id.prototype.sequence = function(f, of) {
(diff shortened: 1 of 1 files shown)


id.js
```diff
@@ -33,8 +33,8 @@ Id.prototype.ap = function(b) {
 };
 
 // Traversable
-Id.prototype.sequence = function(f, of) {
-    return f(this.value).map(function(y){ return new Id(y); });
+Id.prototype.sequence = function(of) {
+    return this.value.map(function(y){ return new Id(y); });
```
davidchambers: Let's use `Id.of` in place of `function(y){ return new Id(y); }`. :)

davidchambers:
I was about to open a pull request for this myself, but paused when I realized the `of` parameter is ignored. Can you explain why this is so? Is it necessary for fold-based implementations, but not for single-value containers such as Id?

DrBoolean:
Hi @davidchambers 

I put a quick comment in there for others as well. The `of` must be provided for types which might fail. A good example would be Maybe:

```js
var j = Just([2])
j.sequence(Array.of)
// [Just(2)]

var n = Nothing
n.sequence(Array.of)
// [Nothing]
```

In the second case, if we didn't provide `Array.of`, we wouldn't know what type to return.

Most instances do not need to take advantage of the `of` argument, but it must be provided to keep a standard api.

davidchambers:
Thanks for the explanation. :)

:deciduous_tree:

PR #118, opened 2016-01-15
Set logo width/height to prevent jumps of #hash links

rpominov:
When someone comes to the project page via a #hash link they can end up at a wrong position. Because  after image loaded page height changes at scroll moves. This PR should fix that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 (aka "Algebraic JavaScript Specification")
 
-![](logo.png)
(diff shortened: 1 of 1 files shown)


davidchambers:
Thank you, thank you, thank you!

This has been bugging me (and confusing me) for years. I'd assumed there must be a quirky bug in GitHub's code. It's wonderful to have this fixed!

rpominov:
Glad that helps. It was bugging me for some time too.

Issue #92, opened 2015-04-26
Change the spec so it would require prefixed method names


rpominov:
I propose to change the spec so it would require more unique method names. For example, instead of `ap` method, a `@@fantasy-land/ap`. It can be called as `foo['@@fantasy-land/ap']()`.

This would serve two purposes: 
  1. It will be easier to add spec support to an existing API, as method names will be guaranteed not taken, 
  2. we could use duck-typing to check if an object implements a spec.

This also will allow to create a polyfill that adds fantasy-land support to native JS data structures. It relatively ok to add `@@fantasy-land/ap` method to `Array.prototype`, but less ok to add `ap`.

Also libraries like Bacon, for example, will be able to add fantasy-land support, as for now it problematic with [map](https://github.com/baconjs/bacon.js/#observable-map) method, for instance, as it not strictly compatible with the spec, but they could add `@@fantasy-land/map` that is.

I think this change could significantly speed up the adoption of fantasy-land specification.

------
The idea inspired by [transducers protocol](https://github.com/cognitect-labs/transducers-js#transformer-protocol), seems like it worked out for them pretty well.

See also discussion on Gitter: https://gitter.im/fantasyland/fantasy-land?at=553cb82d20328f114ca36f0f

CrossEye:
:+1: 

This would make life much easier for [Ramda][ra] which is a generic functional programming library for JS, but which will work well with types implementing the FantasyLand spec:

    R.map(square, [1, 2, 3]); //=> [1, 4, 9]
    R.map(square, Just(5)); //=> Just(5).map(square); //=> Just(25)
    R.map(square, Nothing); //=> Nothing.map(square); //=> Nothing
    R.map(square, Right(7)); //=> Right(7).map(square); //=> Right(49)
    // etc.

While we do this sort of dynamic dispatching on a number of our functions, there is a special place in our hearts for the FantasyLand specification, and I would love to see this work more generally than it does now.  We could easily code to dispatch to such FL versions first if they're defined and after that to matching named function.  This would mean that Ramda might integrate better with Bacon and other libraries that have conflicts with the FL names.

  [ra]: https://github.com/ramda/ramda

SimonRichardson:
Why not just put the fantasy-land library in a wrapper and expose your new
method names. I just don't feel fantasy-land should do this IMO.

So :-1: from me.

CrossEye:
The problem is that FantasyLand want to claim ownership of all of these names:

 - equals
 - concat
[...]

PR #120, opened 2016-01-17
Add index.js that exposes constants for method names

rpominov:
Implementation of idea discussed here https://github.com/fantasyland/fantasy-land/issues/92#issuecomment-172357516

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,63 @@ implemented and how they can be derived from new methods.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## How to add Fantasy Land compatibility to your library
+
+1. Add `fanatasy-land` package as a peer dependency. Add in your `package.json`:
+
+  ```js
+  {
+    ...
+    "peerDependencies": {
+      "fantasy-land": "*"
+    },
+    ...
+  }
+  ```
+
+2. The `fantasy-land` package exposes method names, you should use them for you Fantasy Land methods:
+
+  ```js
+  var fl = require('fanatasy-land')
+
+  // ...
+
+  MyType.prototype[fl.map] = function(fn) {
+    // Here goes implementation of map for your type...
+  }
+  ```
+
+## How to use Fantasy Land compatible library in your application
+
+1. Add library npm package, and `fanatasy-land` as your normal (or dev) dependecies:
+
+  ```js
+  {
+    ...
+    "dependencies": {
+      "some-fl-compatible-lib": "1.0.0",
+      "fantasy-land": "1.0.0"
+    },
+    ...
+  }
+  ```
+
+2. If you don't want to access Fantasy Land methods directly
+  (for example if you use two libraries that talk to each other using Fantasy Land),
+  then that's it — simply install them and use as you normally would,
+  only install `fanatasy-land` package as well.
+
+  If you do want to access Fantasy Land methods, do it like this:
+
+  ```js
+  var fl = require('fanatasy-land')
+  var Something = require('some-fl-compatible-lib')
+
+  var foo = new Something(1)
+  var bar = foo[fl.map](x => x + 1)
+  ```
+
+
 ## Algebras
 
 ### Setoid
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -0,0 +1,13 @@
...
+  empty: 'empty',
+  map: 'map',
+  ap: 'ap',
+  of: 'of',
+  reduce: 'reduce',
+  sequence: 'sequence',
+  chain: 'chain',
+  extend: 'extend',
+  extract: 'extract'
+}
```
SimonRichardson: Did we not want to go the full symbols route or at least try it so we can see if it works or not?

rpominov: This is supposed to be the first backward-compatibility version, that simply uses current names and doesn't prevent name clashes. All current libs will be automatically compatible with this version, I've wrote about this in https://github.com/fantasyland/fantasy-land/issues/92#issuecomment-172357516

Then we can release a second version that uses unique names. I suggest to use strings (something like `fantasy-land/map`) though, really don't see any advantages of using Symbols, and there are number of disadvantages as @robotlolita pointed out in https://github.com/fantasyland/fantasy-land/issues/92#issuecomment-139907109

SimonRichardson: :+1: will have a play tomorrow

davidchambers:
This is a simple, pragmatic solution to a very real problem. Well done!

I suggest not touching symbols with a barge pole.

README.md
```diff
@@ -45,6 +45,63 @@ implemented and how they can be derived from new methods.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## How to add Fantasy Land compatibility to your library
+
+1. Add `fanatasy-land` package as a peer dependency. Add in your `package.json`:
```
davidchambers: Since we're using string values, could one use a regular dependency rather than a peer dependency?

README.md
```diff
@@ -45,6 +45,63 @@ implemented and how they can be derived from new methods.
...
+  // ...
+
+  MyType.prototype[fl.map] = function(fn) {
+    // Here goes implementation of map for your type...
+  }
+  ```
+
+## How to use Fantasy Land compatible library in your application
+
+1. Add library npm package, and `fanatasy-land` as your normal (or dev) dependecies:
```
davidchambers: I suggest removing the reference to dev dependencies, as it won't often be correct to make fantasy-land a dev dependency.

id.js
```diff
@@ -1,60 +1,62 @@
+var fl = require('./index.js')
```
davidchambers: We could abbreviate the import path to `'./'` or `'./index'`.

package.json
```diff
@@ -24,5 +24,6 @@
   },
   "files": [
     "id.js"
-  ]
+  ],
+  "main": "index.js",
```
davidchambers: This is the default value, though it doesn't hurt to be explicit.

README.md
rpominov: I don't think so. For instance if we release two versions `1.0` with not prefixed names and `2.0` with `"fantasy-land/map"` like names. And a library will have a normal dependency  `fantasy-land@2.0.0`, we won't be able to use that library with an older library that uses unprefixed names. 

And with peer dependencies, we can kind of inject `fantasy-land@1.0.0` to that library.

README.md
rpominov: This is for case when FL compatible libs also dev dependencies. Although it's not clear from the sentence, I agree. Yeah, probably easier to just remove that.

package.json
```diff
@@ -24,5 +24,6 @@
   },
   "files": [
     "id.js"
```
SimonRichardson: If we're publishing to npm, what do we think about also including the README.md (the specification) along with it?

rpominov: Docs say README is always included by default https://docs.npmjs.com/files/package.json#files

I wonder if we should add `index.js` here though.

SimonRichardson:
:+1:

rpominov:
Someone who has access should publish this to npm now, so we could start migrating our libs.

rpominov:
@puffnfresh @joneshf ping. What do you think about releasing this thing to npm?

puffnfresh:
Sounds good.

rpominov:
@puffnfresh I was hoping you or @joneshf could actually do this, since you have access to npm. I can submit a PR with version bump if that will help.

SimonRichardson:
Or if someone gave me push access then I can do it for people.

puffnfresh:
@SimonRichardson I think you have write access to this repo and collaborator access on npm. Do you need something else?

SimonRichardson:
I thought I didn't have access to npm, if I do, sorry for my mistake.

PR #121, opened 2016-01-18
Fix derived Foldable.toArray

XeCycle:
Array.prototype.concat flattens the argument if it is an array, so
concat with a single-element array here.

Hope I am not mistaken?


davidchambers:
LGTM

PR #123, opened 2016-01-27
Add the laws implementations

SimonRichardson:
The motivation behind this is to allow any person wanting to implement
any structure that should conform to the specification will also have
the readme and more importantly the implementations of the laws.

The advantage of this as well is that any changes to the spec should be
reflected in the laws folder.

### DIFF
--- a/laws/applicative.js
+++ b/laws/applicative.js
@@ -1,0 +1,24 @@
+'use strict';
+
+const {identity, apply} = require('fantasy-combinators');
+const {ap} = require('../index');
+
+const id = (t) => (eq) => (x) => {
+    const a = t.of(identity).ap(t.of(x));
+    const b = t.of(x);
+    return eq(a, b);
+};
+
+const homomorphism = (t) => (eq) => (x) => {
+    const a = t.of(identity).ap(t.of(x));
+    const b = t.of(identity(x));
+    return eq(a, b);
+};
+
+const interchange = (t) => (eq) => (x) => {
+    const a = t.of(identity).ap(t.of(x));
+    const b = t.of(apply(t)).ap(t.of(x));
+    return eq(a, b);
+};
+
+modules.exports = { identity: id, homomorphism, interchange };
--- a/laws/applicative.js
+++ b/laws/applicative.js
@@ -1,24 +1,39 @@
 'use strict';
 
-const {identity, apply} = require('fantasy-combinators');
-const {ap} = require('../index');
+const {identity, thrush} = require('fantasy-combinators');
+const {of, ap} = require('..');
 
-const id = (t) => (eq) => (x) => {
-    const a = t.of(identity).ap(t.of(x));
-    const b = t.of(x);
+/**
+
+### Applicative
+
+1. `a.of(x => x).ap(v)` is equivalent to `v` (identity)
+2. `a.of(f).ap(a.of(x))` is equivalent to `a.of(f(x))` (homomorphism)
+3. `u.ap(a.of(y))` is equivalent to `a.of(f => f(y)).ap(u)` (interchange)
+
+**/
+
+const identityʹ = t => eq => x => {
+    const a = t[of](identity).ap(t[of](x));
+    const b = t[of](x);
     return eq(a, b);
 };
 
-const homomorphism = (t) => (eq) => (x) => {
-    const a = t.of(identity).ap(t.of(x));
-    const b = t.of(identity(x));
+const homomorphism = t => eq => x => {
+    const a = t[of](identity).ap(t[of](x));
+    const b = t[of](identity(x));
     return eq(a, b);
 };
 
-const interchange = (t) => (eq) => (x) => {
-    const a = t.of(identity).ap(t.of(x));
-    const b = t.of(apply(t)).ap(t.of(x));
+const interchange = t => eq => x => {
+    const u = t[of](identity);
+
+    const a = u.ap(t[of](x));
+    const b = t[of](thrush(x)).ap(u);
     return eq(a, b);
 };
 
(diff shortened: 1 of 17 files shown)


SimonRichardson:
Note: not all laws implemented yet, more to come, but wondered what people thought about this PR?

laws/applicative.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const {identity, apply} = require('fantasy-combinators');
+const {ap} = require('../index');
```
davidchambers: `require('..')` is sufficient here. :)

laws/applicative.js
```diff
@@ -0,0 +1,24 @@
...
+
+const id = (t) => (eq) => (x) => {
```
davidchambers: These parens are all optional.

laws/applicative.js
```diff
@@ -0,0 +1,24 @@
...
+    return eq(a, b);
+};
+
+const interchange = (t) => (eq) => (x) => {
+    const a = t.of(identity).ap(t.of(x));
+    const b = t.of(apply(t)).ap(t.of(x));
+    return eq(a, b);
+};
+
+modules.exports = { identity: id, homomorphism, interchange };
```
davidchambers: [`\n`](https://github.com/editorconfig/editorconfig/wiki/Newline-at-End-of-File-Support)?

davidchambers:
Including these files is a great idea!

laws/applicative.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const {identity, apply} = require('fantasy-combinators');
+const {ap} = require('../index');
+
+const id = (t) => (eq) => (x) => {
+    const a = t.of(identity).ap(t.of(x));
```
rpominov: This should be `t[of](identity)[ap](t[of](x))` I think. 

Btw we have a problem with `of` here, it's a reserved word, we can't use it for variable name :crying_cat_face:

SimonRichardson: Didn't know that! OK will change

laws/applicative.js
SimonRichardson: Learned something new today :dancer:

laws/applicative.js
SimonRichardson: Yeah, I'm still under sided if they should be required for consistency. I'll remove for now.

laws/applicative.js
SimonRichardson: not fussed either way.

joneshf:
This is a great idea for most of the laws, but some might require more thought. For instance, the composition law for `Functor`. It will not suss out a bad `Functor` all the time. I know, these are tests, not proofs so you can't find all the issues, but take the typical "maybe" implementation that springs up fairly often:

```js
function NullHelper(x) {
  return {
    map: f => x == null ? NullHelper(x) : NullHelper(f(x)),
  };
}
```

While it passes both the tests given, that's definitely not a `Functor`.

rpominov:
Hm, why we allow specify arbitrary `x`, btw? Can we check laws with our own values, like `0, 1, null, undefined` (or event do quick-check)? Is there cases when a type can work only with limited set of values?

SimonRichardson:
@rpominov The idea is I want to get quick check to call these, so we need to pass in values.

rpominov:
Quick check should catch @joneshf's example then, I think.

SimonRichardson:
fyi: i've moved this discussion to the gitter room, if anybody wants to join me?

SimonRichardson:
@DrBoolean I'm having a problem getting the `Id` implementation of `sequence` to validate without changing the identity law.

The specification requires this:
```
2. `u.map(x => Id(x)).sequence(Id.of)` is equivalent to `Id.of` (identity)
```
Is it not?
```
2. `u.map(identity).sequence(Id.of)` is equivalent to `Id.of` (identity)
```

Or am I missing what `u` should be?
https://github.com/fantasyland/fantasy-land/blob/0540c4206665ea3f041a556ce1fa461bdc3fdde2/id_test.js#L98

SimonRichardson:
I've also implemented `id_test` so that we can verify that the laws pass with `Id`. We can also pass this through travis to make sure that nothing broke between spec updates.

rpominov:
I think this law should be:

```
2. `u.map(x => Id(x)).sequence(Id.of)` is equivalent to `Id.of(u)` (identity)
```

So, for example, `u = [1,2,3]`, we can either do this:

```js
u.map(Id) // [Id(1),Id(2),Id(3)]
  .sequence(Id.of) // Id([1,2,3])
```

Or just:

```js
Id.of(u) // Id([1,2,3])
```

SimonRichardson:
Should `Id` be able to sequence over itself?

SimonRichardson:
@rpominov so I got it to pass, it's just a bit weird that `Id` can't sequence over itself.

DrBoolean:
Hi!

@rpominov is right. `u` is another functor and I accidentally wrote the `Id.of` pointfree which is confusing and not even equivalent.

I think the confusion is that in `sequence(Id.of)`, the `Id.of` part is ignored

DrBoolean:
The reason for the argument is for failure. So we're not sequencing Id over itself, but rather supplying type help.

`Some(x).sequence(Id.of) == Id(Some(x))`
vs
`None.sequence(Id.of) == Id(None)`

SimonRichardson:
@DrBoolean makes sense once I've implemented it :dancer: Everybody happy for me to merge?

DrBoolean:
Heck yeah! Want me to add that `Id.of(u)` to the spec real quick?

SimonRichardson:
Yes please. Just add it to this branch.

Issue #128, opened 2016-03-08
Traversable definition uses undefined terms


CrossEye:
I'm bothered by the definition of [`Traversable`][tr].  The remaining specifications use plain English, depend on one another, and build on "value" and "equivalent" from the [Terminology][te] section.

But `Traversable` uses three terms not covered in any of these places:

  * natural transformation
  * `Id`
  * `Compose`

My least favorite course in grad school was Category Theory, and I remember little of it, but it means that when I run across this stuff 25 years later, it's at least familiar, and I have some intuitive feel for what makes something a natural transformation.  But intuitive feel is not a good basis for a specification.

And it's pretty easy to guess at what `Id` and `Compose` should be and how they should work.  But there should be no guessing in a spec.

Unfortunately, I don't feel a strong enough grasp of Traversable to offer a reasonable alternative, and I apologize.  But I still think it's worth pointing out the issue.
 

  [te]: https://github.com/fantasyland/fantasy-land#terminology
  [tr]: https://github.com/fantasyland/fantasy-land#traversable

rpominov:
It would be great to have natural transformation definition in terms of laws a JS function must obey in order to be natural transformation.

I'm trying to grasp it by reading https://en.m.wikipedia.org/wiki/Natural_transformation but not there yet.

rpominov:
As for `Id` and `Compose` maybe we can refer to implementations js files in this repo? We don't have one for Compose though, but I think I could make a PR.

rpominov:
Hm, isn't natural transformation a pair of functions and not one function?

![image](https://cloud.githubusercontent.com/assets/825702/13845699/b512ff4e-ec52-11e5-95a9-7decd971709d.png)

For example, given our functors are Array and Maybe, we need a function that can convert `[] -> Maybe` and another one that converts `Maybe -> []`. And those two functions must obey the 'naturality square' law.

Is that correct?

What confuses me is that in Traversable law only one function is used. How a one function alone can be natural transformation?

rpominov:
I tried to express law for natural transformation in js, is it correct?

```
t(f.of(x)).map(a) ≡ t(f.of(x).map(a))
```

CrossEye:
> As for `Id` and `Compose` maybe we can refer to implementations js files in this repo.

I believe that a specification should not refer to implementations.  Moreover, I think it would be better to reconstrue this in terms of plain functions rather than these otherwise unmotivated types.

rpominov:
[...]

PR #68, opened 2014-05-29
Foldable / Traversable

DrBoolean:
Not sure if everything is how we want it yet, but thought i'd create this pull request to make it easier to see the changes.

PR #131, opened 2016-03-27
Fix traversable laws

rpominov:
See #128

Opening the PR to get this moving.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,8 +261,24 @@ for any `t` such that `t(x).map(a)` is equivalent to `t(x.map(a))` (naturality)
 
 2. `u.map(F.of).sequence(F.of)` is equivalent to `F.of(u)` for any Applicative `F` (identity)
 
-3. `u.map(Compose).sequence(Compose.of)` is equivalent to
-   `Compose(u.sequence(f.of).map(x => x.sequence(g.of)))` (composition)
+3. `u.map(Compose.of).sequence(Compose.of)` is equivalent to
+   `Compose.of(u.sequence(f.of).map(x => x.sequence(g.of)))` for Compose type defined below (composition)
+
+```js
+Compose = {
+  of(c) {
+    return {
+      c,
+      map(f) {
+        return Compose.of(c.map(y => y.map(f)))
+      },
+      ap(x) {
+        return Compose.of(c.map(u => y => u.ap(y)).ap(x.c))
+      }
+    }
+  }
+}
+```
 
 * `traverse`; derivable as `function(f, of) { return this.map(f).sequence(of); }`
 
--- a/README.md
+++ b/README.md
@@ -265,19 +265,21 @@ for any `t` such that `t(x).map(a)` is equivalent to `t(x.map(a))` (naturality)
    `Compose.of(u.sequence(f.of).map(x => x.sequence(g.of)))` for Compose type defined below (composition)
 
 ```js
-Compose = {
-  of(c) {
-    return {
-      c,
-      map(f) {
-        return Compose.of(c.map(y => y.map(f)))
-      },
-      ap(x) {
-        return Compose.of(c.map(u => y => u.ap(y)).ap(x.c))
-      }
-    }
-  }
-}
+var Compose = function(c) {
+  this.c = c;
+};
+
+Compose.of = function(c) {
+  return new Compose(c);
+};
+
+Compose.prototype.ap = function(x) {
+  return Compose.of(this.c.map(u => y => u.ap(y)).ap(x.c));
+};
+
+Compose.prototype.map = function(f) {
+  return Compose.of(this.c.map(y => y.map(f)));
+};
 ```
 
 * `traverse`; derivable as `function(f, of) { return this.map(f).sequence(of); }`
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ A value that implements the Traversable specification must also
 implement the Functor specification.
 
 1. `t(u.sequence(f.of))` is equivalent to `u.map(t).sequence(g.of)`
-where `t` is a natural transformation from `f` to `g` (naturality)
+for any `t` such that `t(x).map(a)` is equivalent to `t(x.map(a))` (naturality)
 
 2. `u.map(x => Id(x)).sequence(Id.of)` is equivalent to `Id.of(u)` (identity)
 
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ implement the Functor specification.
 1. `t(u.sequence(f.of))` is equivalent to `u.map(t).sequence(g.of)`
 for any `t` such that `t(x).map(a)` is equivalent to `t(x.map(a))` (naturality)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -257,12 +257,28 @@ A value that implements the Traversable specification must also
 implement the Functor specification.
 
 1. `t(u.sequence(f.of))` is equivalent to `u.map(t).sequence(g.of)`
-where `t` is a natural transformation from `f` to `g` (naturality)
-
-2. `u.map(x => Id(x)).sequence(Id.of)` is equivalent to `Id.of(u)` (identity)
-
-3. `u.map(Compose).sequence(Compose.of)` is equivalent to
-   `Compose(u.sequence(f.of).map(x => x.sequence(g.of)))` (composition)
+for any `t` such that `t(x).map(a)` is equivalent to `t(x.map(a))` (naturality)
```
rpominov: It's not clear here what `x` is. Maybe we should rewrite it as `t(f.of(x)).map(a) ≡ t(f.of(x).map(a))`?

README.md
```diff
@@ -257,12 +257,28 @@ A value that implements the Traversable specification must also
...
+
+2. `u.map(F.of).sequence(F.of)` is equivalent to `F.of(u)` for any Applicative `F` (identity)
```
rpominov: Are we sure this works for any Applicative and not only Id?

README.md
```diff
@@ -257,12 +257,28 @@ A value that implements the Traversable specification must also
...
+      c,
+      map(f) {
+        return Compose.of(c.map(y => y.map(f)))
+      },
+      ap(x) {
+        return Compose.of(c.map(u => y => u.ap(y)).ap(x.c))
+      }
+    }
+  }
+}
```
rpominov: Is this code style ok for this code? Do we want to use ES6? Should we rewrite it with `prototype` or `ES6 class` maybe?

I also not 100% sure this implementation is correct. Will double check later.

SimonRichardson: Prototype, as that's how `Id` is implemented, we should at least be
consistent.

rpominov:
Rewrote Compose via `.prototype` and checked that it works in Ramda repl: http://goo.gl/ExPdoo

Also GitHub for some reason hided previous comments on other parts of the diff, although I didn't touch those parts: 
image
These comments still need attention.

SimonRichardson:
@rpominov 👍  on the prototype stuff.

CrossEye:
This looks great!  Thanks.

joneshf:
While we're in here fixing this. Can we change the requirement to `traverse` rather than `sequence`? With `traverse` you can derive `map` and `reduce` for free. But you can't do that with `sequence`.

joneshf:
Oh, actually. I guess we don't even have the requirement that `Traversable` implement `Foldable` as well. Better add that while fixing this as well.

rpominov:
> While we're in here fixing this. Can we change the requirement to traverse rather than sequence? 

We could, but that would be a braking change, do we really want that?

> With traverse you can derive map and reduce for free. But you can't do that with sequence.

We can derive `traverse` using `sequence` and `map`, and then we can derive `reduce`. So we only missing possibility of deriving `map`, `reduce` still can be derived.

On other hand using `traverse` will be consistent with Monad that uses `chain` and not `join`.

Should we add instructions how `reduce` can be derived? It's kind of tricky and requires two helper types `Const` and `Endo` from what I understood from https://github.com/fantasyland/fantasy-land/pull/68#issuecomment-44789091 . Do we want that in the spec?

I'm very curious what @DrBoolean thinks about all this, since he originally brought Traversable and Foldable to the spec.

DrBoolean:
Love that you got rid of `Id`! I was thinking since `Id` was the example in the spec, it'd be okay to reference, but no one, including myself, wanted to mix in implementation. I guess there still is implementation there with `Compose`,  but it's better inline.

A traversable is a foldable functor, yes, but I don't see the reason to enforce foldable. I think it should just be known that one can derive it. (https://github.com/fantasyland/fantasy-land/pull/68#issuecomment-44789091)

I'm cool with switching to `traverse`, I originally wrote (ported) it that way, but found it to be more complex and only defined that way for edge case performance reasons. That said, i hardly ever use `sequence` and reach for `traverse` myself.

rpominov:
> Love that you got rid of Id! I was thinking since Id was the example in the spec, it'd be okay to reference, but no one, including myself, wanted to mix in implementation.

Great, we wasn't sure if we can replace Id with any Applicative in #128

> A traversable is a foldable functor, yes, but I don't see the reason to enforce foldable.

Seems like in other places in the spec when we can derive one algebra from another that algebra is enforced.

> I'm cool with switching to `traverse`

+1 from me, and I can add this to PR, but still not sure if everybody want this.

DrBoolean:
As far as any applicative goes. It'd be interesting to search for a counter example. Perhaps `Const` will break it

rpominov:
> Perhaps Const will break it

No, it works with `Const`. `u.map(Const.of).sequence(Const.of)` will create a bunch of `Const(empty())` which then will be concatenated (using `ap`) to a single `Const(empty())`; and `Const.of(u)` also will give us `Const(empty())`.

DrBoolean:
word!

DrBoolean:
As far as "breaking change", its since traverse and sequence can be defined in terms of each other, I don't see it as much of an issue. Especially if we provide derived sequence: `traverse(F.of, id)`

rpominov:
Made Traversable to depend on Foldable, and added derived `reduce()` (checked that it works here http://goo.gl/nW1bkB)

DrBoolean:
I disagree that we should say it must implement the foldable spec. Even though something is definitely foldable if it's traversable, we don't depend on `reduce` and it confuses the work involved to make a type traversable. One absolutely must have `map`, but not `fold` or `reduce`

I feel providing a derivable version is enough to show it's foldable.

rpominov:
I don't have a strong opinion on this, but I think we should try to keep spec consistent. As I understand, before Traversable and Foldable were added, the one of rules to decide where a dependency must be introduced was: if algebra B can be derived from algebra A, then A must depend on B. An example of such case is `Chain` depending on `Apply`.

This seems relevant to #134 and #132

rpominov:
Maybe it would be better to replace `sequence` with `traverse` in another PR, so it wouldn't block other changes from this PR? I can open PR for `traverse` right after this is merged.

If that ok, the open questions left here:
 - Should we add the requirement that `Traversable` implement `Foldable`? (currently added)
 - Is the derived `reduce()` I wrote fine?

DrBoolean:
I'm good with this. The derived `traverse` does seem out of place now that we've talked it all out, but i suppose that is a separate PR.

I have to say, amazing job on derived `reduce` and the naturality law!

I think this is a terrific clarification of the spec. Thanks for all your work on this.

Everyone else cool to update this?

SimonRichardson:
Looks good to me.

rpominov:
Great! The PR that replaces `sequence` with `traverse` is coming soon.

CrossEye:
:sparkles:

PR #134, opened 2016-04-12
derivations do not make certain methods optional

### LINKED ISSUE #132: derivable methods are contradictory
[Applicative](https://github.com/fantasyland/fantasy-land#applicative):

> A value that implements the Applicative specification must also implement the Apply specification.

[Apply](https://github.com/fantasyland/fantasy-land#apply):

> A value that implements the Apply specification must also implement the Functor specification.

[Functor](https://github.com/fantasyland/fantasy-land#functor):

> A value which has a Functor must provide a `map` method.

So a value that implements the Applicative specification must provide a `map` method.

Here is the contradiction:

> A value which satisfies the specification of an Applicative does not need to implement:
>
>   - Functor's `map`; derivable as `function(f) { return this.of(f).ap(this); }`

Is there a semantic difference between "implements the Applicative specification" and "satisfies the specification of an Applicative"?

I'm wondering whether Ramda's `map` function should operate on a value which provides `of` and `ap` but not `map`. The fact that derivations are mentioned in the spec suggests they are desirable, but as written it doesn't seem correct to derive implementations.

davidchambers:
Closes #132

Commit message:

> Derivations of FL methods are relevant to implementers of data types.
>
> Derivations of FL methods are irrelevant to implementers of functions which operate on values of these types, since the values must provide the relevant methods.

I'm not sure how `toArray` and `traverse` fit into the specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,7 @@ An algebra is a set of values, a set of operators that it is closed
 under and some laws it must obey.
 
 Each Fantasy Land algebra is a separate specification. An algebra may
-have dependencies on other algebras which must be implemented. An
-algebra may also state other algebra methods which do not need to be
-implemented and how they can be derived from new methods.
+have dependencies on other algebras which must be implemented.
 
 ## Terminology
 
@@ -100,7 +98,6 @@ implemented and how they can be derived from new methods.
   var foo = new Something(1)
   var bar = foo[fl.map](x => x + 1)
   ```
-
 
 ## Algebras
 
@@ -209,11 +206,6 @@ method takes one argument:
 A value that implements the Applicative specification must also
 implement the Apply specification.
 
-A value which satisfies the specification of an Applicative does not
-need to implement:
-
-* Functor's `map`; derivable as `function(f) { return this.of(f).ap(this); }`
-
 1. `a.of(x => x).ap(v)` is equivalent to `v` (identity)
 2. `a.of(f).ap(a.of(x))` is equivalent to `a.of(f(x))` (homomorphism)
 3. `u.ap(a.of(y))` is equivalent to `a.of(f => f(y)).ap(u)` (interchange)
@@ -232,9 +224,7 @@ or its `constructor` object. The `of` method takes one argument:
 
 ### Foldable
 
-1. `u.reduce` is equivalent to `u.toArray().reduce`
-
-* `toArray`; derivable as `function() { return this.reduce((acc, x) => acc.concat([x]), []); }`
+1. `u.reduce` is equivalent to `u.reduce((acc, x) => acc.concat([x]), []).reduce`
 
 #### `reduce` method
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,9 +30,7 @@ An algebra is a set of values, a set of operators that it is closed
 under and some laws it must obey.
 
 Each Fantasy Land algebra is a separate specification. An algebra may
-have dependencies on other algebras which must be implemented. An
-algebra may also state other algebra methods which do not need to be
-implemented and how they can be derived from new methods.
```
joneshf: I didn't realize this sentence existed in the spec. That makes most of what I said in #132 questionable.

rpominov:
While we at it, maybe we should add something like:

> If a method can be derived and the type also has a hand written version of that method, derived and hand written methods must be equivalent. And if two derived version of a method can be created (for instance we can derive `map` using `ap` or `chain`), they must be equivalent.

To clarify that matter once and for all.

See Gitter discussion: https://gitter.im/fantasyland/fantasy-land?at=570a28a4c65c9a6f7f284c1c

rpominov:
> I'm not sure how toArray and traverse fit into the specification.

I also don't quite understand why they are here. I think `toArray` is necessary to define `reduce`, it restraints how it can be implemented. But `traverse` seems not needed at all.

davidchambers:
> I think `toArray` is necessary to define `reduce`, it restraints how it can be implemented.

That may be the case, but the wording can surely be improved. Currently we state:

> 1.  `u.reduce` is equivalent to `u.toArray().reduce`
>
> - `toArray`; derivable as `function() { return this.reduce((acc, x) => acc.concat([x]), []); }`

Could we remove mention of `toArray` by inlining?

> 1.  `u.reduce` is equivalent to `u.reduce((acc, x) => acc.concat([x]), []).reduce`

DrBoolean:
I like that `toArray` illustrates the power of foldable in a nutshell and gives a law associated with the algebra. I agree that `traverse` is unnecessary. Perhaps if traversable is rewritten in terms of `traverse`, it will justify deriving `sequence` for legacy reasons.

README.md
```diff
@@ -393,12 +372,25 @@ The `extract` method takes no arguments:
...
+  - [`map`][] may be derived from [`chain`][] and [`of`][]:
 
+        function(f) { var m = this; return m.chain(a => m.of(f(a))); }
 
+  - [`ap`][] may be derived from [`chain`][]:
 
+        function(m) { return this.chain(f => m.map(f)); }
+
+If a data type provides a method which *could* be derived, its behaviour must
+be equivalent to that of the derivation (or derivations).
```
davidchambers: I added this sentence based on your suggestion, @rpominov.

davidchambers:
The `:+1:` in https://github.com/fantasyland/fantasy-land/issues/132#issuecomment-219176464 indicates that @puffnfresh is in favour of this change, so let's merge it. :)

I've reverted the removal of `toArray` based on feedback from @rpominov and @DrBoolean.

README.md
rpominov: Great!

rpominov:
We should also move `reduce` derivation that was added after this PR was opened.

rpominov:
> I've reverted the removal of toArray based on feedback from @rpominov and @DrBoolean.

I like your idea of replacing it with a law:

> 1. `u.reduce` is equivalent to `u.reduce((acc, x) => acc.concat([x]), []).reduce`

davidchambers:
Thank you for the feedback, @rpominov. I've made the following changes:

  - removed mention of `toArray` by inlining;
  - moved the `reduce` derivation to the Derivations section; and
  - added syntax highlighting to derivations.

Avaq:
I prefer this version of the spec over the current. :+1:

puffnfresh:
:+1:

davidchambers:
Thanks, @puffnfresh! Can we merge this now? @SimonRichardson?

SimonRichardson:
Sure, do it

davidchambers:
I'm hoping someone with write access will hit the merge button. ;)

SimonRichardson:
Haha, done!

PR #135, opened 2016-04-17
Use traverse() as base method for Traversable

rpominov:
See #131

Summary:
 - We can derive `map` using `traverse`.
 - To use `traverse` in Traversable is more consistent with using `chain` and not `join` in Monad.

Derived methods in Ramda REPL: http://goo.gl/PdLOYk


davidchambers:
Could you resolve the merge conflicts, @rpominov?

rpominov:
Yep, done.

SimonRichardson:
@rpominov can you update this too the new master please.

rpominov:
Done. And travis seems to work :)

README.md
```diff
@@ -277,18 +277,25 @@ Compose.prototype.map = function(f) {
 };
 ```
 
-#### `sequence` method
+#### `traverse` method
 
 ```hs
-sequence :: Apply f, Traversable t => t (f a) ~> (b -> f b) -> f (t a)
+traverse :: Apply f, Traversable t => t a ~> ((a -> f b), (c -> f c)) -> f (t b)
```
davidchambers: Should this be `Applicative f` rather than `Apply f`? `c -> f c` is `of` essentially, isn't it?

davidchambers: I asked the question [here][1] too, since the type signature was added in #147.


[1]: https://github.com/fantasyland/fantasy-land/pull/147/files#r79309177

README.md
```diff
@@ -251,13 +251,13 @@ method takes two arguments:
 A value that implements the Traversable specification must also
 implement the Functor and Foldable specifications.
 
-1. `t(u.sequence(f.of))` is equivalent to `u.map(t).sequence(g.of)`
+1. `t(u.traverse(x => x, F.of))` is equivalent to `u.traverse(t, G.of)`
 for any `t` such that `t(a).map(f)` is equivalent to `t(a.map(f))` (naturality)
```
davidchambers: Could someone explain this to me? Unless I missed something `F` and `G` have not been introduced by this point. I see that they are mentioned below.

I struggle to see how an expression involving `F` (but not `G`) can be equivalent to an expression involving `G` (but not `F`). How is this possible?

rpominov: The signature of `t` here is something like `t :: Applicative F, Applicative G => G a -> F a`. So both `F` and `G` are present on both sides via `t`.

I agree this is not very obvious though, one need to derive signature of `t` by looking at the expressions in order to understand this.

rpominov: Or rather not `t :: Applicative F, Applicative G => F a -> G a` but `t :: F a -> G a`, in other words `t` works not with any two Applicatives but specifically with types `F` and `G`.

Maybe the law should say "... for any `t :: F a -> G a` such ...".

davidchambers: Thanks, @rpominov. It seems this law could be expressed more clearly, but I'm not the person to rework it. ;)

PR #136, opened 2016-05-05
Fix for monad identity laws.

jwoudenberg:
I believe there is an issue with the implementation of the identity monad laws. Specifically, the identity function is not a valid argument to `chain`, because it will not return a value of the same Monad (a requirement inherited from Chain).
The reason this has not caused an error so far is because is an implementation detail of the Monad it is tested against. I believe other valid monads can run into issues when testing them against the law in it's current state.

My proposed fix is passing in a monad-returning function `f` to test against instead of using `identity`. I realize this is not such a nice solution in the sense that it changes the API of these tests with regards of the laws of the other tests, so a better solution would be welcome.

The only trivial monad-returning function we could use is `t.of`, but using that would result in identitical `leftIdentity` and `rightIdentity` tests. Another approach is to build a simple `f` in the law file itself, but it would require knowledge of the type of `x`, which would compromise the generality of the test.

### DIFF
--- a/id_test.js
+++ b/id_test.js
@@ -34,6 +34,9 @@ Sum.prototype[equals] = function(x) {
     return this.v.equals ? this.v.equals(x.v) : this.v === x.v;
 };
 
+// Special function for monad identity laws.
+const f = x => Id.of(x + x);
+
 const equality = (x, y) => x.equals ? x.equals(y) : x === y;
 const test = f => t => {
     t.ok(f("x"));
@@ -74,8 +77,8 @@ exports.functor = {
 };
 
 exports.monad = { 
-    leftIdentity: test((x) => monad.leftIdentity(Id)(equality)(x)),
-    rightIdentity: test((x) => monad.rightIdentity(Id)(equality)(x))
+    leftIdentity: test((x) => monad.leftIdentity(Id)(equality)(x)(f)),
+    rightIdentity: test((x) => monad.rightIdentity(Id)(equality)(x)(f))
 };
 
 exports.monoid = { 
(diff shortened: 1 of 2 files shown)


id_test.js
```diff
@@ -34,6 +34,9 @@ Sum.prototype[equals] = function(x) {
     return this.v.equals ? this.v.equals(x.v) : this.v === x.v;
 };
 
+// Special function for monad identity laws.
+const f = x => Id.of(x + x);
+
```
puffnfresh: What's the `+` for?

jwoudenberg: Without it, `x => Id.of(x)` is simply the same as `Id.of`. In that case, there is no difference between the `leftIdentity` and `rightIdentity` test cases. So I opted for a simple non-trivial monad-returning function.

jwoudenberg:
I found another issue in that the [Monoid laws](https://github.com/fantasyland/fantasy-land/blob/master/laws/monad.js) currently assume the Monoid implements an `of` method, which is not required. Is there an interesting in fixing these? If so, I'll happily add a suggestion to this PR :).

gabejohnson:
@jwoudenberg you're linking to the Monad laws there.

jwoudenberg:
@gabejohnson Thanks for pointing that out to me. The point remains though, see [this new and improved link](https://github.com/fantasyland/fantasy-land/blob/master/laws/monoid.js#L16).

SimonRichardson:
PR would be amazing😉

jwoudenberg:
I'll see what I can do :).

SimonRichardson:
This is fixed in #171

PR #137, opened 2016-05-07
Profunctor and bifunctor specification

scott-christopher:
A slight variation of the profunctor and bifunctor specs raised in #124, removing the mention of `lmap` and renaming `dimap` to `promap`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ structures:
 * [Extend](#extend)
 * [Comonad](#comonad)
 * [Bifunctor](#bifunctor)
+* [Profunctor](#profunctor)
 
 <img src="figures/dependencies.png" width="677" height="212" />
 
@@ -376,6 +377,34 @@ method takes two arguments:
 
 3. `bimap` must return a value of the same Bifunctor.
 
+### Profunctor
+
+A value that implements the Profunctor specification must also implement
+the Functor specification.
+
+1. `p.promap(a => a, b => b)` is equivalent to `p` (identity)
+2. `p.promap(a => f(g(a)), b => h(i(b)))` is equivalent to `p.promap(f, i).promap(g, h)` (composition)
+
+#### `promap` method
+
+A value which has a Profunctor must provide a `promap` method.
+
+The `profunctor` method takes two arguments:
+
+    c.promap(f, g)
+
+1. `f` must be a function which returns a value
+
+    1. If `f` is not a function, the behaviour of `promap` is unspecified.
+    2. `f` can return any value.
+
+2. `g` must be a function which returns a value
+  
+    1. If `g` is not a function, the behaviour of `promap` is unspecified.
+    2. `g` can return any value.
+
+3. `promap` must return a value of the same Profunctor
+
 ## Derivations
 
 When creating data types which satisfy multiple algebras, authors may choose
@@ -392,11 +421,17 @@ to implement certain methods then derive the remaining methods. Derivations:
     ```js
     function(f) { var m = this; return m.chain(a => m.of(f(a))); }
     ```
- 
+
   - [`map`][] may be derived from [`bimap`]:
   
     ```js
     function(f) { return this.bimap(a => a, f); }
+    ```
+
+  - [`map`][] may be derived from [`promap`]:
+
+    ```js
+    function(f) { return this.promap(a => a, f); }
     ```
 
   - [`ap`][] may be derived from [`chain`][]:
@@ -450,5 +485,6 @@ be equivalent to that of the derivation (or derivations).
 [`extract`]: #extract-method
 [`map`]: #map-method
 [`of`]: #of-method
+[`promap`]: #promap-method
 [`reduce`]: #reduce-method
 [`sequence`]: #sequence-method
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ structures:
 * [Monad](#monad)
 * [Extend](#extend)
 * [Comonad](#comonad)
+* [Bifunctor](#bifunctor)
 
 <img src="figures/dependencies.png" width="677" height="212" />
 
@@ -348,6 +349,33 @@ The `extract` method takes no arguments:
 1. `extract` must return a value of type `v`, for some variable `v` contained in `w`.
     1. `v` must have the same type that `f` returns in `extend`.
 
+### Bifunctor
+
+A value that implements the Bifunctor specification must also implement
+the Functor specification.
+
+1. `p.bimap(a => a, b => b)` is equivalent to `p` (identity)
+2. `p.bimap(a => f(g(a)), b => h(i(b))` is equivalent to `p.bimap(g, i).bimap(f, h)` (composition)
+
+#### `bimap` method
+
+A value which has a Bifunctor must provide an `bimap` method. The `bimap`
+method takes two arguments:
+
+    c.bimap(f, g)
+
+1. `f` must be a function which returns a value
+
+    1. If `f` is not a function, the behaviour of `bimap` is unspecified.
+    2. `f` can return any value.
+
+2. `g` must be a function which returns a value
+
+    1. If `g` is not a function, the behaviour of `bimap` is unspecified.
+    2. `g` can return any value.
+
+3. `bimap` must return a value of the same Bifunctor.
+
 ## Derivations
 
 When creating data types which satisfy multiple algebras, authors may choose
@@ -363,6 +391,12 @@ to implement certain methods then derive the remaining methods. Derivations:
 
     ```js
     function(f) { var m = this; return m.chain(a => m.of(f(a))); }
+    ```
+ 
+  - [`map`][] may be derived from [`bimap`]:
+  
+    ```js
+    function(f) { return this.bimap(a => a, f); }
     ```
 
   - [`ap`][] may be derived from [`chain`][]:
@@ -407,6 +441,7 @@ be equivalent to that of the derivation (or derivations).
 
 
 [`ap`]: #ap-method
+[`bimap`]: #bimap-method
 [`chain`]: #chain-method
 [`concat`]: #concat-method
 [`empty`]: #empty-method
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -394,10 +396,57 @@ The `extract` method takes no arguments:
...
+
+    1. If `g` is not a function, the behaviour of `bimap` is unspecified.
+
+
+### Profunctor
+
+A value that implements the Profunctor specification must also implement
+the Functor specification. The Profunctor specification is a contravariant
+functor on its first type parameter and covariant functor on its second type
+parameter.
```
Avaq: > must also implement the Functor specification

Why is this the case for `Profunctor`, but not (explicitly) for `Bifunctor`?

README.md
```diff
@@ -394,10 +396,57 @@ The `extract` method takes no arguments:
...
+
+1. `f` must be a function which returns a value
+
+    1. If `f` is not a function, the behaviour of `promap` is unspecified.
+    2. `f` can return any value.
+
+2. `g` must be a function which returns a value
+  
+    1. If `g` is not a function, the behaviour of `promap` is unspecified.
+    2. `g` can return any value.
```
Avaq: > `f` can return any value
> `g` can return any value

These rules are explicitly stated for `promap` but not for `bimap`, yet the spec for `bimap` makes no mention of the types of values its given functions may return.

README.md
```diff
@@ -394,10 +396,57 @@ The `extract` method takes no arguments:
     1. `v` must have the same type that `f` returns in `extend`.
 
 
+### Bifunctor
 
+A value that implements the Bifunctor specification is a functor with two
+type arguments that are both covariant.
```
rpominov: Do we need this sentence? Can we simply omit it, like in Functor? For the sake of consistency.

README.md
```diff
@@ -394,10 +396,57 @@ The `extract` method takes no arguments:
...
+#### `bimap` method
 
+A value which has a Bifunctor must provide an `bimap` method. The `bimap`
+method takes two arguments:
+
+    c.bimap(f, g)
+
+1. `f` must be a function which returns a value
+
+    1. If `f` is not a function, the behaviour of `bimap` is unspecified.
```
rpominov: We need here "`f` can return any value." and "`bimap` must return a value of the same Bifunctor	" below for consistency with Functor.

README.md
```diff
@@ -394,10 +396,57 @@ The `extract` method takes no arguments:
...
+
+2. `g` must be a function which returns a value
+
+    1. If `g` is not a function, the behaviour of `bimap` is unspecified.
+
+
+### Profunctor
+
+A value that implements the Profunctor specification must also implement
+the Functor specification. The Profunctor specification is a contravariant
```
rpominov: Can we not use "contravariant" and "covariant" terms here and replace it with laws if necessary? And if this don't add any restrictions maybe just omit it.

rpominov:
Will it make sense to also add derivation of `map`?

README.md
scott-christopher: It's probably not necessary given that it is mentioned in the second sentence now. I'm happy to remove it.

README.md
scott-christopher: I think it is necessary two indicate that there are two type variables that are modified by `bimap`. We could potentially drop the mention of covariance though.

README.md
scott-christopher: Mentioning `contravariance` here is intended to highlight the differences between `promap` and `bimap`. I'd gets difficult to convey in words without the help of some kind of type signature, but I'm open to suggestions.

README.md
scott-christopher: If we want people to be able to assume that they can call `map` on a profunctor or bifunctor, then we should add "must also implement the Functor specification" to the bifunctor spec as well. If not, we should remove it here for consistency.

README.md
scott-christopher: Yeah, I don't really see much value in stating this. Happy to remove them.

scott-christopher:
> Will it make sense to also add derivation of map?

If we add the restriction that both Bifunctor and Profunctor require Functor, then I think it would be helpful to add the derivations of `map` for each.

README.md
```diff
@@ -394,10 +396,57 @@ The `extract` method takes no arguments:
...
+functor on its first type parameter and covariant functor on its second type
+parameter.
+
+1. `p.promap(a => a, b => b)` is equivalent to `p` (identity)
+2. `p.promap(compose(f1)(f2), compose(g1)(g2))` is equivalent to `p.promap(f1, g1).promap(f2, g2)` (composition)
```
rpominov: The order of composition seems wrong. Should be:

```js
p.promap(compose(f1)(f2), compose(g2)(g1)) ≡ p.promap(f1, g1).promap(f2, g2)
```

Also we probably should use "inlined" function composition like in the rest of the spec, and I would use four letters instead of letters with numbers, so:

```js
p.promap(x => f(g(x)), x => h(i(x))) ≡ p.promap(f, i).promap(g, h)
```

I took letters from https://hackage.haskell.org/package/profunctors-5.2/docs/Data-Profunctor.html

README.md
rpominov: It's just my opinion but I think spec should focus on restrictions, explanations of types should live elsewhere. So if it's a restriction it should go into laws in some form. But I think it's more like an explanation.

README.md
rpominov: I think this difference is expressed as the difference in "composition" law of two types.

Btw I also think that it would be better to use type signatures in the spec instead of writing down types in english, how it's currently done.

rpominov:
+1 from me for adding dependency on Functor and derivation rules of map to both types. This way we also make sure that if a type has hand written `map` it's compatible with Profunctor/Bifunctor.

rpominov:
Sorry for all the nitpicking, I trying to be consistency nazi. Hopefully this helps :)

README.md
scott-christopher: Good catch. :zap:

README.md
scott-christopher: I've removed the mention of `contravariance` from the spec.

README.md
scott-christopher: :zap:

README.md
scott-christopher: I have removed this sentence given there is no mention of type arguments elsewhere in the spec.

README.md
scott-christopher: I've left them in there given they exist elsewhere in the spec though I've since realised this is actually incorrect for `f`, as it must return a type of the profunctor's first type argument. The problem is that we don't mention type arguments elsewhere in the spec so I'm not sure how we could best state this requirement. Again, a type signature would be useful here.

README.md
```diff
@@ -394,10 +396,71 @@ The `extract` method takes no arguments:
...
+
+### Profunctor
+
+A value that implements the Profunctor specification must also implement
+the Functor specification.
+
+A value which satisfies the specification of a Profunctor does not need to
+implement:
+
+* Functor's `map`; derivable as `f => this.promap(a => a, f)`
```
rpominov: `this` works differently in arrow functions, we need to use `function` here.

Otherwise everything looks good to me now.

scott-christopher: :zap:

Avaq:
I believe this PR would have to be rewritten a little to comply with the changes made in #134. Specifically, the derivatives need to be moved down to their own chapter.

davidchambers:
As @Avaq pointed out, this pull request needs a few tweaks.

scott-christopher:
I'll try to find some time to get it up to date this week.

scott-christopher:
I've pushed up some changes to make it consistent with #134.

README.md
```diff
@@ -365,6 +422,18 @@ to implement certain methods then derive the remaining methods. Derivations:
...
+  - [`map`][] may be derived from [`bimap`]:
+  
+    ```js
+    function(f) { return this.bimap(a => a, f); }
+    ```
+
+  - [`map`][] may be derived from [`promap`]:
+
+    ```js
+    function(f) { return this.promap(a => a, f) }
```
davidchambers: We're missing a semicolon after the closing paren.

laws/bifunctor.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+### Bifunctor
+
+1. `p.bimap(a =>, b => b)` is equivalent to `p` (identity)
+2. `p.bimap(compose(f1)(f2), compose(g1)(g2))` is equivalent to `p.bimap(f1, g1).bimap(f2, g2)` (composition)
+
+**/
+
+const identityʹ = t => eq => x => {
```
davidchambers: We can use primes in identifiers now? Neat!

scott-christopher: I believe it has been possible since ES5, as it belongs to the [`Modifier letter (Lm)`](https://codepoints.net/U+02B9) category.

README.md
scott-christopher: Order has been restored :D :zap:

davidchambers:
I plan to define these methods for various data types. I'd love to see this pull request merged, as it's nice to have documentation point to the FL spec rather than an unmerged pull request. :)

scott-christopher:
Any other objections from this being merged?

ping @SimonRichardson, @joneshf

SimonRichardson:
:+1:

joneshf:
Thanks so much!

davidchambers:
:tada:

index.js
```diff
@@ -9,5 +9,7 @@ module.exports = {
   sequence: 'sequence',
   chain: 'chain',
   extend: 'extend',
-  extract: 'extract'
+  extract: 'extract',
+  bimap: 'bimap',
```
rpominov: @SimonRichardson Could you release a new version on NPM that would include this changes to `index.js`?

SimonRichardson: https://www.npmjs.com/package/fantasy-land @0.3.0

rpominov: Thanks!

Issue #144, opened 2016-07-25
Changing ap arguments order for consistency with map and chain, and for better types support


rpominov:
The current order of arguments (if we count `this` as an argument) in `ap` is problematic. I see two problems about it.

### It's inconsistent with `map` and `chain`

```
map :: (this: A, f: A => B) => MyType<B>
chain :: (this: A, f: A => MyType<B>) => MyType<B>
ap :: (this: MyType<A => B>, x: MyType<A>) => MyType<B>

// sorry about made up types lang, I hope it's clear what I mean here
```

In `map` and `chain` value goes first and function goes second, in `ap` the other way around.

### It's impossible to type with Flow and probably other type systems

... or at least hard, and I didn't figured out how to do it.

```js
// @flow

// With current arguments order we have to use `any`, and therefore loose type info
class MyType<A> {
  ap(other: MyType<any>): MyType<any>
}

// With reversed order it's straightforward
class MyType<A> {
  ap<B>(other: MyType<A => B>): MyType<B>
}
```

I understand this is a braking change that will probably break a lot of code. But I just want to open this discussion, and hear what others think. Maybe this is a pain point for many people, and it would be reasonable to make the change.

SimonRichardson:
It's been mentioned before, I'll dig out the references, but I'm in favour as well.

SimonRichardson:
https://github.com/fantasyland/fantasy-land/issues/50#issuecomment-27708812

rpominov:
Ah, right. #50 is specifically about this problem. I remember it was mentioned before, but thought that it was mentioned briefly in an unrelated issue. I guess I created a duplicate then, sorry. I'll close this one.

SimonRichardson:
Let's do it, open a PR.

rpominov:
OK, I'll open one tomorrow.

Issue #122, opened 2016-01-20
Namespaced prefixes — next step for resolving name clashes


rpominov:
So #92 was somewhat resolved with this PR #120

The next step is to eventually publish a version of `fantasy-land` npm package with namespaced prefixed names.

There are some questions to discuss though:

 - When do we release the new version? Should we wait until majority of libs migrate to the new approach, or should we release it right away so two versions will coexist allowing everyone to choose on which versions to depend in each particular case?
 - How the prefixed names should look like?

Maybe there more questions, these two is what came to my mind.

rpominov:
I've faced a couple of problems with the new approach. Not sure where better to document them, hopefully this issue is fine.

1. `of` is a reserved word, we can't use it for a variable name. Workarounds are:
  
  ```diff
  - import {map, of} from 'fantasy-land'
  + import {map, of as just} from 'fantasy-land'
  ```
  ```diff
  - const {map, of} = require('fantasy-land')
  + const {map, of: just} = require('fantasy-land')
  ```
  ```js
  // Alternatively just import whole object to a variable
  import fl from 'fantasy-land'
  foo[fl.of](1)
  ```
  Maybe we could add an alias for `of` in index.js (e.g. `just`)

2. [Flow](http://flowtype.org/) doesn't support dynamic names for methods (neither via `class` syntax nor via `.prototype`):
  ```
  src/fantasy.js:34
   34:   [concat]( other:Stream<T> ): Stream<T> {
          ^ computed property keys not supported
  ```
  ```
  src/fantasy.js:147
  147: Stream.prototype[concat] = Stream.prototype.concat
       ^^^^^^^^^^^^^^^^^^^^^^^^ assignment of computed property/element. Indexable signature not found in
  147: Stream.prototype[concat] = Stream.prototype.concat
       ^^^^^^ Stream
  ```
  A workaround I found so far looks like this:
  ```js
  addFLmethods(Stream.prototype)

  function addFLmethods( proto:any ): void {
    proto[concat] = proto.concat
  }
  ```
  But this is only a workaround for adding methods to a class, in any case we bypass Flow type checking when will use those class instances. And in general dynamic names can be a problem for any static analysis tools like Flow/TypeScript/etc.

rpominov:
I was thinking, maybe in the future we could start treating all this as a temporal measure for the transition period and go back to using hardcoded names but with prefixes.

rpominov:
[...]

Issue #50, opened 2013-10-26
Question about Applicatives


cwmyers:
Hi guys,

I've been implementing FantasyLand in monet.js and I've come a little unstuck with `ap()`.

Unless I've got the wrong end of the stick FantasyLand's `ap()` demands that the applicative already contains a function and then accepts an applicative with a value.  

`appWithFunction.ap(appWithValue)`

Unfortunately I've followed scalaz & functional java approach which does the opposite.  

`appWithValue.ap(appWithFunction)`

See here for the `Maybe` example: http://cwmyers.github.io/monet.js/#apmaybefn

Here is an example from FunctionalJava `Option`:
`public final <B> Option<B> apply(Option<F<A,B>> of)` 

So my question is why has Fantasy Land taken this particular approach to applicative?  I'm worried that this would be a large breaking change to the monet library and I'm trying to figure out what the right thing is to do.

Cheers,
Chris

SimonRichardson:
Should we flip the order of `ap`, or just live with it?

rpominov:
I'm having trouble typing `ap()` with Flow, ended up with this:

```js
  // Apply
  ap<A,B>( other:Stream<A> ): Stream<B> {
    // In order for .ap() to work `this` must be a `Stream<A => B>`,
    // but I don't know how to express that constraint in Flow, so just use `any`
    const streamF:BasicStream<any> = this.observe
    return new Stream(bs.ap(streamF)(other.observe))
  }
```

Would be straightforward if it was other way around.

joneshf:
That is quite unfortunate. Can you actually write the other way with flow?

rpominov:
Sure, with reversed version of ap it will look like this:

```js
class Stream<T> {

  ap<A>( other:Stream<( x:T ) => A> ): Stream<A> {
    return new Stream(bs.ap(other.observe)(this.observe))
  }

}
```

rpominov:
For the record: not a big problem for me, just wanted to add my two cents :)

SimonRichardson:
I've been hit with this recently as well. So the question is how do we change it, because I think we should!

Could we maybe rename `ap` to something else, or do we say if you depend on the new fantasy-land #92 `ap` works like this now?

rpominov:
We could certainly use NPM to help us manage changes like that after libraries start depending on FL.

We could do for example:

 - `1.0` backward compatibility release with unprefixed names and no changes to the spec
[...]

PR #145, opened 2016-07-26
Flip arguments order in ap

rpominov:
See #50 and #144

This is a breaking change and we have two options how to do it:

 1. Like any other changes that were done before (except they weren't breaking)
 2. Add prefixes to method names (#122) and bump major version number

If we go with the first option there will be a period when some libraries already updated but some not, so they might not work well with each other. For example a code like this may stop working:

```js
import R from 'ramda'
import Foo from 'foo'

a = R.lift(x => y => x + y, Foo.of(1), Foo.of(2))
```

In this case users of libraries will have to freeze their dependencies' versions until all libs are updated.

If we do the second, a user of the libraries will be able to force one that already updated to work in `v0.2` mode, by installing `fantasy-land@0.2.1` instead of `fantasy-land@1.0.0`. But the library that updated will need to support this too, by doing something like:

```js
import fl from 'fantasy-land'

const lift2 = fl.version // supposedly we add .version property in 1.0
  ? (fn, x, y) => x[fl.map](x => y => fn(x, y))[fl.ap](y)
  : (fn, x, y) => y[fl.ap](x[fl.map](x => y => fn(x, y)))

export default lift2
```

Currently PR implements the first option, and I'll be happy to update it to the second if we decide so. And I, for one, think that we should go with the second option.

Whatever we decide, I think we should wait for some time before merging so everybody will have a chance to discuss.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,12 +175,12 @@ method takes one argument:
 A value that implements the Apply specification must also
 implement the Functor specification.
 
-1. `a.map(f => g => x => f(g(x))).ap(u).ap(v)` is equivalent to `a.ap(u.ap(v))` (composition)
+1. `v.ap(u.ap(a.map(f => g => x => f(g(x)))))` is equivalent to `v.ap(u).ap(a)` (composition)
 
 #### `ap` method
 
 ```hs
-ap :: Apply f => f (a -> b) ~> f a -> f b
+ap :: Apply f => f a ~> f (a -> b) -> f b
 ```
 
 A value which has an Apply must provide an `ap` method. The `ap`
@@ -188,24 +188,24 @@ method takes one argument:
 
     a.ap(b)
 
-1. `a` must be an Apply of a function,
+1. `b` must be an Apply of a function,
 
-    1. If `a` does not represent a function, the behaviour of `ap` is
+    1. If `b` does not represent a function, the behaviour of `ap` is
        unspecified.
 
-2. `b` must be an Apply of any value
+2. `a` must be an Apply of any value
 
-3. `ap` must apply the function in Apply `a` to the value in
-   Apply `b`
+3. `ap` must apply the function in Apply `b` to the value in
+   Apply `a`
 
 ### Applicative
 
 A value that implements the Applicative specification must also
 implement the Apply specification.
 
-1. `a.of(x => x).ap(v)` is equivalent to `v` (identity)
-2. `a.of(f).ap(a.of(x))` is equivalent to `a.of(f(x))` (homomorphism)
-3. `u.ap(a.of(y))` is equivalent to `a.of(f => f(y)).ap(u)` (interchange)
+1. `v.ap(a.of(x => x))` is equivalent to `v` (identity)
+2. `a.of(x).ap(a.of(f))` is equivalent to `a.of(f(x))` (homomorphism)
+3. `a.of(y).ap(u)` is equivalent to `u.ap(a.of(f => f(y)))` (interchange)
 
 #### `of` method
 
@@ -268,8 +268,8 @@ Compose.of = function(x) {
   return new Compose(F.of(G.of(x)));
 };
 
-Compose.prototype.ap = function(x) {
-  return new Compose(this.c.map(u => y => u.ap(y)).ap(x.c));
+Compose.prototype.ap = function(f) {
+  return new Compose(this.c.ap(f.c.map(u => y => y.ap(u))));
 };
 
 Compose.prototype.map = function(f) {
@@ -470,13 +470,13 @@ to implement certain methods then derive the remaining methods. Derivations:
   - [`map`][] may be derived from [`ap`][] and [`of`][]:
 
     ```js
-    function(f) { return this.of(f).ap(this); }
+    function(f) { return this.ap(this.of(f)); }
     ```
 
   - [`map`][] may be derived from [`chain`][] and [`of`][]:
 
     ```js
-    function(f) { var m = this; return m.chain(a => m.of(f(a))); }
+    function(f) { return this.chain(a => this.of(f(a))); }
     ```
 
   - [`map`][] may be derived from [`bimap`]:
@@ -494,7 +494,7 @@ to implement certain methods then derive the remaining methods. Derivations:
   - [`ap`][] may be derived from [`chain`][]:
 
     ```js
-    function(m) { return this.chain(f => m.map(f)); }
+    function(m) { return m.chain(f => this.map(f)); }
     ```
 
   - [`reduce`][] may be derived as follows:
@@ -511,7 +511,7 @@ to implement certain methods then derive the remaining methods. Derivations:
         return this;
       };
       Const.prototype.ap = function(b) {
-        return new Const(f(this.value, b.value));
+        return new Const(f(b.value, this.value));
       };
       return this.map(x => new Const(x)).sequence(Const.of).value;
     }
--- a/id.js
+++ b/id.js
@@ -36,7 +36,7 @@ Id.prototype[fl.map] = function(f) {
 
 // Apply
 Id.prototype[fl.ap] = function(b) {
-    return new Id(this.value(b.value));
+    return new Id(b.value(this.value));
 };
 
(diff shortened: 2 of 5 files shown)


SimonRichardson:
:+1:

SimonRichardson:
We should try and merge this one in tbh!

@joneshf thoughts?

scott-christopher:
Does your second option require all FL types in scope to implement the same version?

The only way I can see this landing without causing pretty major headaches would be to include a version in the method namespace, such as:
```
```

rpominov:
I was thinking about separate name for flipped ap too, this is an options yes. Basically we can do detection as `'fl@2/ap' in someApply` or as `fl.version > 1` (or just `fl.version` if we add `version` only starting from `1.0`).

But actually all this solutions seem so complicated, maybe it's indeed better to just break some code for some period of time. It is hard for me to estimate amount of trouble this can make. I, for one, maintain very little of code that will need to be updated. So for me it's not a big issue. But I guess some people would rather do something like second option because of bigger amount of code they maintain.

scott-christopher:
If we were to go with the `const { version } = require('fantasy-land')` approach and stick with keeping it as a peer dependency, wouldn't that require both the implementers of the spec and consumers of the spec to handle the different version ranges (unless I've misunderstood)?

e.g.
```js
const { ap, version } = require('fantasy-land')

Id.prototype[ap] = version >= 2 ? (other) => new Id(other.value(this.value))
                                : (other) => new Id(this.value(other.value))

//:: Apply f => f (a -> b) -> f a -> f b
const ap = (ap1, ap2) => version >= 2 ? ap2[ap](ap1)
                                      : ap1[ap](ap2)
```

Personally I'm not so sure that this change offers enough to warrant the break in backwards compatibility and added complexity.

joneshf:
> We should try and merge this one in tbh!
> 
> @joneshf thoughts?

I'm almost definitely a YES! This has always confused me. Let me read this more thoroughly first though. :)

rpominov:
@scott-christopher Yea, both implementers and consumers would need to do the detection. Pretty  complicated stuff, I agree.

SimonRichardson:
I'd rather make this a breaking change, that's why we implemented semver. So I personally don't mind bumping to `1.0.0`, that's what it's therefore.

davidchambers:
> If we were to go with the `const { version } = require('fantasy-land')` approach and stick with keeping it as a peer dependency, wouldn't that require both the implementers of the spec and consumers of the spec to handle the different version ranges (unless I've misunderstood)?

It wouldn't *require* this, as far as I can tell. The implementor of the Id data type could release a version which is only guaranteed to work with `fantasy-land@1.x.x` (supporting older versions as well would be quite confusing, I think). Authors of libraries which provide functions that operate on values of FL-compatible data types have the *choice* of supporting both `0.3.x` and `1.x.x`.

---

Side note: In Node, at least, there's already a way to access the package's version:

```javascript
require('fantasy-land/package.json').version
```

SimonRichardson:
> Authors of libraries which provide functions that operate on values of FL-compatible data types have the choice of supporting both 0.3.x and 1.x.x.

Bingo!

scott-christopher:
I guess I'm not so concerned whether this is a backwards breaking change as much as there is a sane way for users and libraries interacting with FL types to manage these breaking changes going forward.

> Authors of libraries which provide functions that operate on values of FL-compatible data types have the choice of supporting both 0.3.x and 1.x.x.

I apologise for not following along here, but could someone spell out how a library could support one type that implements the `0.3.x` version of `ap` and another that supports `1.x` at the same time? Or is there a constraint imposed on the user of such a library to pick one version that all types and supporting libraries must support in their application?

SimonRichardson:
> could someone spell out how a library could support one type that implements the 0.3.x version of ap and another that supports 1.x at the same time

To put it bluntly we can't, but we shouldn't leave things broken just because of that.

davidchambers:
> [C]ould someone spell out how a library could support one type that implements the `0.3.x` version of `ap` and another that supports `1.x` at the same time?

The library would need to provide two versions of each `ap`-related function, I think. In Ramda's case, the library could provide both `R.ap` and `R._ap`.

> Or is there a constraint imposed on the user of such a library to pick one version that all types and supporting libraries must support in their application?

It needn't be all or nothing, but the burden would be on the user to know with which version of FL each of her data types is compatible, and choose—for example—`R.ap` or `R._ap` as appropriate.

This would, though, undermine one of the main benefits of FL: the ability to define functions which operate on many data types *without specific knowledge of these data types*.

If we're to make this change, I don't foresee a seamless transition. It seems the best approach would be to update the `ap` implementations for all the data types as quickly as possible, then release a new version of Ramda compatible with these updated data types.

scott-christopher:
> It seems the best approach would be to update the ap implementations for all the data types as quickly as possible, then release a new version of Ramda compatible with these updated data types.

Right, but that's my concern. This change would introduce a state where libraries such as Ramda could no longer offer interoperability between types where some FL-implementing libraries update sooner than others. Given that interoperability is one of the main selling points of FL, I guess I'm failing to see how the current state is considered broken compared to the state that this could potentially introduce.

I may also be in the minority in thinking that there's nothing particularly wrong with the current spec, which may be influencing my thinking here. IMO, the existing spec feels familiar to the both `Function.prototype.apply` and `<*>`.

I don't want to stop movement on this if most people feel like this is the right decision. I'd just like to know that there is a way to manage the transition. If there is no easy way to manage the transition, I'd like to know that others feel this is important enough to cause the potential disruption.

rpominov:
Just to add two cents:

> This change would introduce a state where libraries such as Ramda could no longer offer interoperability between types where some FL-implementing libraries update sooner than others.

Users could just stick with older versions for a period of time until all libraries they use together are updated.

> I'd like to know that others feel this is important enough to cause the potential disruption.

To me the main selling point is better support for type systems like Flow https://github.com/fantasyland/fantasy-land/issues/50#issuecomment-173400402

SimonRichardson:
> Given that interoperability is one of the main selling points of FL

For that version.

If we're never allowed to make breaking changes (ignoring adding items to the specification) then this project is dead! That means we can never learn from our mistakes, or removing dead weight and update the spec accordingly. The idea of "fantasy-land" is born from the fact that the promise specification wouldn't make the sacrifice of learning and updating when confronted with new knowledge.

scott-christopher:
Sure, like I said I don't have an issue with breaking changes, my point is that we look for ways of managing the transition and upgrade for those that depend on FL through the breaking changes, otherwise we're potentially just pushing the inability to move forward onto the users.

SimonRichardson:
Cool 👍 

Either we should burden the library writers aka https://github.com/fantasyland/fantasy-land/pull/145#issuecomment-237731165 or we expose `ap'` for sometime whilst people understand that a transition is being made.

README.md
```diff
@@ -413,17 +413,17 @@ to implement certain methods then derive the remaining methods. Derivations:
...
     ```js
-    function(f) { return this.of(f).ap(this); }
+    function(f) { return this.ap(this.of(f)); }
     ```
 
   - [`map`][] may be derived from [`chain`][] and [`of`][]:
 
     ```js
-    function(f) { var m = this; return m.chain(a => m.of(f(a))); }
+    function(f) { return this.chain(a => this.of(f(a))); }
```
joneshf: :100:

SimonRichardson:
Shall we merge this, I think we should.

rpominov:
If #146 will happen, would be great to release both of these PRs in the same version.

SimonRichardson:
I don't mind merging this, and holding the release until 146 lands.

SimonRichardson:
Considering #146 will be merged, can we update to the latest master please.

rpominov:
I'll update it after #146 lands if that's ok. It just would be easier to me as this and #146 also might have conflicts with each other.

SimonRichardson:
@rpominov This is missing the new `chainRec` specification.

SimonRichardson:
#146 has landed.

rpominov:
⚡

davidchambers:
:tada:

Issue #149, opened 2016-08-21
Contravariant Functor?


rjmk:
Copied from https://github.com/rpominov/static-land/issues/15

> Would [fantasyland] be interested in a contravariant functor added to the spec?

> It would have a method contramap :: Contravariant f => (a -> b, f b) -> f a. It would become a dependency to profunctor. The laws are pretty much the same as for functor, but with a flip on the functions in composition

> Happy to make a PR for [people] to look over

SimonRichardson:
:+1:

rpominov:
Probably one of reasons why this didn't happen from the beginning is that we didn't want to claim another name for FL in people's methods namespaces. If that was the case, then this is another argument for #146 .

SimonRichardson:
Closing this as the discussions are now in the PR. Reopen if you think otherwise.

PR #150, opened 2016-08-21
adds contravariant functor to spec

rjmk:
Resolves https://github.com/fantasyland/fantasy-land/issues/149


README.md
```diff
@@ -380,7 +402,7 @@ method takes two arguments:
 ### Profunctor
 
 A value that implements the Profunctor specification must also implement
-the Functor specification.
+the Functor and Contravariant Functor specifications.
```
scott-christopher: I'm in two minds about including this here.

In some other typed languages we'd have an issue where the variance of the type arguments of contravariant and profunctors don't line up. We don't have this issue with a regular covariant functor and profunctor because both last type arguments are covariant (we just have to partially apply the type constructor to take the kind of profunctor from `* -> * -> *` to `* -> *`).

The other half of my mind says this is probably being pedantic as we're not bound by such a type system so it's not really a problem.

laws/contravariant.js
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+const identityʹ = t => eq => x => {
+    const a = t(x)[contramap](identity);
+    const b = t(x);
+    return eq(a, b);
+};
+
+const composition = t => eq => x => {
+    const a = t(x)[contramap](compose(identity)(identity));
```
scott-christopher: I'm wondering whether we should use some contravariant function here rather than identity, as we don't really ensure correct composition order while using `identity`.

Not sure what the best way to generate an arbitrary function from within these tests would be though, other than having some memoised function that returns a randomly generated number e.g.

```js
const fn = () => {
  const received = new Map()
  return x => {
    if (!received.has(x)) received.set(x, Math.random())
    return received.get(x)
  }
}

const composition = t => eq => x => {
    const f = fn();
    const g = fn();
    const a = t(x)[contramap](compose(f)(g));
    const b = t(x)[contramap](f)[contramap](g);
    return a === b;
};
```

I believe the same issue exists in the profunctor test too.

rjmk: @scott-christopher Might be worth bringing in [JSVerify](https://github.com/jsverify/jsverify) if we want to do lots of this sort of thing. Might also be overkill for toy laws

README.md
rjmk: Hmm.

Am I write in thinking that there could be a class `AmateurFunctor` with the opposite problem?

```hs
class AmateurFunctor f where
  amateurmap :: (a -> b) -> (c -> d) -> f a d -> f b c
```

I'm slightly unclear if there's a connection between the variance of the type arguments in the kind and the variance of type parameters in the type.

scott-christopher: > Am I write in thinking that there could be a class AmateurFunctor with the opposite problem?

In something like Haskell, a type that implements `AmateurFunctor`  could only be declared as `Contravariant`.
```hs
class AmateurFunctor f where
  amateurmap :: (a -> b) -> (c -> d) -> f a d -> f b c

class Contravariant f where
  contramap :: (b -> a) -> f a -> f b

instance AmateurFunctor f => Contravariant (f a) where
  contramap = amateurmap id
```

To also declare it as a `Functor` you would need to flip the order of the type arguments with a `newtype`.
```hs
newtype Flip f b a = Flip { unFlip :: f a b }

instance AmateurFunctor f => Functor (Flip f a) where
  fmap f = Flip . amateurmap f id . unFlip
```
If however we were to ignore the notion of type arguments, it could just be stated that:
```js
p.promap(f, g) = p.contramap(f).map(g) = p.map(g).contramap(f)
```
It looks a little odd for me to see `contramap` and `map` being applied directly to the same type, but as the `Flip` newtype above highlights, the types are effectively identical.

laws/contravariant.js
scott-christopher: perhaps an alternative would just be to allow `f` and `g` to be provided as arguments?

README.md
rjmk: > > Am I right in thinking that there could be a class AmateurFunctor with the opposite problem?

> In something like Haskell, a type that implements AmateurFunctor could only be declared as Contravariant.

Yup, that's the sort of thing I was thinking.

I think I get it now. It's for the same reason that the functor for a tuple is on the second element and for a function is on the codomain?

One thing I'm slightly confused about is the variance of type arguments. I think that in a value level function `f :: a -> b -> c`, both of the first two params are in negative position. Is that different for a kind `T :: * -> * -> *`? Or am I misapplying reasoning in an inappropriate domain?

joneshf: I don't think it's that straight forward. Take `Flip f a b`. How these variables are used depends on `f`. If you have

```haskell
newtype Reader a b = Reader (a -> b)
newtype Op a b = Op (b -> a)
```

Then, `Flip Reader a b` has the `a` and `b` in negative and positive position, respectively.
But, `Flip Op a b` has the `a` and `b` in positive and negative position, respectively.
All while `Flip`s kind is still `* -> * -> *`.

There may be some relationship though.

rjmk: I see! I think. Here's a statement my understanding would predict to be true: there's a `Contravariant` instance for `Op`, but no `Functor`. Is that correct?

I'm still not quite sure I get

> In some other typed languages we'd have an issue where the variance of the type arguments of contravariant and profunctors don't line up

In that I think I see the problem now, but I'm not sure I see it's relation to variance of the type arguments

Also there seems to me to be an inessential semantic issue at hand. In that one speaks of, say, the functor instance for functions, but it's seems somewhat conventional / tied to Haskell's syntax that we say fixing the first type variable gives us 'the' functor instance. Might it not be more accurate to talk of the functor instance for functions of fixed domain?

It may be that I have a fairly deep misunderstanding here that would take a lot of teaching effort to resolve, so will happily accept being told to go and read something etc.

joneshf: > Here's a statement my understanding would predict to be true: there's a `Contravariant` instance for `Op`, but no `Functor`. Is that correct?

Yep!

> Might it not be more accurate to talk of the functor instance for functions of fixed domain?

Probably :wink:.

I'll let @scott-christopher respond to your other questions, as the lack of type variables is confusing me.

rjmk: Thanks, @joneshf.

I think I am also pro removing it as a dependency. Are you happy for me to push up the removal, @scott-christopher?

rpominov:
Do we want to add `promap` derivation via `map` and `contramap` to [Derivations][https://github.com/fantasyland/fantasy-land#derivations]. Or do you conclude that this is a bad idea to connect `promap` and `contramap` (the discussion above went way over my head)?

rpominov:
Btw, derivations also act as restrictions (additionally to laws):

> If a data type provides a method which could be derived, its behaviour must be equivalent to that of the derivation (or derivations).

Maybe that adds something to the discussion :)

rjmk:
@rpominov The issue is that a type `P a b`, if there's a profunctor instance for `P` has a (covariant) functor instance for `P a`, but a contravariant functor instance for `P _ b` (which is not a valid expression). So then in fact the derivation could not be achieved

(If we were to allow the 'partial application' in the type to be in the middle as well as the end, then we would not know whether the contrafunctor instance for `P a b` was on the `a` or the `b` so we wouldn't be able to have a reliable derivation)

rpominov:
Ah, I see. So the problem is that Haskell's higher kinded types don't allow this because it doesn't support `P _ b` syntax. But this problem exist only in Haskel, right? And also everything is still correct semantically, the problem is only with syntax.

Do we need to care about this in a specification that targets JavaScript?

rjmk:
I'm not sure it's only syntax (I am far from an expert); I worry that we introduce a possibility for unlawfulness with that degree of freedom

rjmk:
Some more thoughts on this

# Pro

* As it stands, there's nothing to stop someone defining `Semigroup b => instance Monad Either _ b`, for Fantasy Land (please excuse the notation)

# Con

* Part of the purpose of this repo is to help people port knowledge from languages with ADTs
  * Demanding a contravariant instance for a profunctor would be very confusing for them
* We're confident that the way it is in Haskell fits together sensibly. I, at least, am not sure I want to gamble that nothing goes wrong when you stop considering type constructors as kind functions and partially applying them to get the various instances
* Being a dependency to profunctor has little effect on if contravariant is useful

That said, I think let's drop the mention and maybe raise an issue to return to it?

SimonRichardson:
Can you update to the latest master (you'll also need to add a type signature), then it's one step closer to merging.

Thanks.

rjmk:
⚡ 

I also removed the dependency mention

SimonRichardson:
@rjmk you need to update again sorry, more changes have just been dropped. 🎱

rjmk:
⚡

laws/contravariant.js
SimonRichardson: @scott-christopher @rjmk Is this solved, before we release this?

scott-christopher: @SimonRichardson This can be addressed outside of this PR, as it affects the profunctor spec too.

rpominov:
Should this thing be called Cofunctor for consistency with Comonad?

rpominov:
Probably not. Cofunctor is the same as Functor http://stackoverflow.com/questions/34732571/why-there-is-no-cofunctor-typeclass-in-haskell/34732721

rjmk:
You may be interested to know that there is no contravariant (co)monad! Because if you tried to nest the instances of the monad, you'd have contra-contravariance, which is just covariance! So your join would be unlawful

Another way at looking at it is that the contravariance ruins the monoidality of the endofunctor ;)

rpominov:
Lecture about Contravariant Functor from Category Theory course by Bartosz Milewski: https://youtu.be/wtIKd8AhJOc?t=26m46s .

It can be interesting for people who are still learning (like me). I like the analogy he makes that if Functor is a box, than Contravariant Functor is like an engine that needs a value of type `a` as fuel in order to run. Creates good intuition!

Also he talks a bit about the [name (Cofunctor vs Contravariant Functor)](https://www.youtube.com/watch?v=wtIKd8AhJOc&feature=youtu.be&t=33m27s) and about connection between [Contravariant Functor and Profunctor](https://www.youtube.com/watch?v=wtIKd8AhJOc&feature=youtu.be&t=45m29s).

rjmk:
How do people feel about this?

safareli:
I think`figures/` should be updated too.

README.md
```diff
@@ -186,6 +187,31 @@ method takes one argument:
...
+### Contravariant Functor
+
+1. `u.contramap(a => a)` is equivalent to `u` (identity)
+2. `u.contramap(x => f(g(x)))` is equivalent to `u.contramap(f).contramap(g)`
+(composition)
+
+#### `contramap` method
+
+```hs
+contramap :: Contravariant f => f a ~> (b -> a) -> f b
```
davidchambers: If `Contravariant` is the name of the type class, should we remove "Functor" from the heading?

evilsoft: Are there any examples of contra variance on value based ADTs?

I have only seen them on function based bits that have their returns fixed. I know it moves away from the Haskell definition, but would it make sense to adopt a signature more like:

`Contravariant f => f a b ~> (c -> a) -> f c b`

or

`Contravariant f => f (a -> b) ~> (c -> a) -> f (c -> b)`

Or is that to limiting for the spec?

evilsoft: Or....does this just imply that the return is fixed and only specifies where the variance occurs?

In any case, it may make for an easier intuition for consumers determining if they indeed have some contra variance available to them

joneshf: Sure, you can have something like.

```haskell
data Foo a = Foo
```

And that can be contravariant:

```haskell
instance Contravariant Foo where
  contramap _ _ = Foo
```

Interestingly, it's also covariant :smile: 

```haskell
instance Functor Foo where
  fmap _ _ = Foo
```

Or if you're lazy:

```haskell
{-# LANGUAGE DeriveFunctor #-}

data Foo a = Foo deriving (Functor)
```

We shouldn't use the signatures above because they do limit the spec too much. For instance `Foo` couldn't be `Contravariant` with either of those signatures.

evilsoft: Perfect! Thanks @joneshf!! just the answer I was looking for!!

safareli:
and size of img tag in readme should also be updated

JeffDownie:
Is it a good idea to hide the dependency of profunctor? A profunctor looks (at least to my eyes) a lot harder to understand than a contravariant functor anyway - if someone can figure out their object a profunctor, they can figure out is a contravariant functor.

Hiding the links between these structures really detracts from using fantasyland as a learning resource too.

rjmk:
In Haskell, there is no dependency of Profunctor on Contravariant functor. Contravariant functors only have one 'free' type variable, whereas Profunctors have two. For any profunctor `P`, we can define `newtype P' b a = P' (P a b)` such that `P' b` is a contravariant functor (nb: *not* `P'`). Considering that we use Haskell syntax for the type signatures, I think we should probably stick to what they can express

JeffDownie:
It just seems weird, though, limiting ourselves to what a different languages' type signature can represent.

Haskell doesn't seem like the perfect standard to hold this repo to (at least in this case, don't get me wrong, most of the time it's amazing :D). Saying something is or isn't a given mathematical structure based on the order of arguments in its' constructor just feels icky/wrong.

rjmk:
I get that concern. However, considering the construction of the rest of the spec in terms of Haskell syntax, I don't think this PR is the appropriate place to diverge

JeffDownie:
True - that sound excessive for this PR :D

README.md
joneshf: To answer @davidchambers question: yes, I think we should rename this to `Contravariant`. Afterwhich, we should merge this.

@rjmk Are you up to this last change? If not, let me know and I'll make it and get this merged.

rjmk: :zap:

README.md
```diff
@@ -4,7 +4,7 @@
 
 (aka "Algebraic JavaScript Specification")
 
-<img src="logo.png" width="200" height="200" />
+<img src="logo.png" width="200" height="271" />
```
davidchambers: Is this change intentional? Perhaps you meant to update the dimensions of __figures/dependencies.png__ instead?

rjmk: Oh dear! Yep, that's right

figures/dependencies.dot
```diff
@@ -13,6 +13,7 @@ digraph {
   Extend;
   Foldable;
   Functor;
+  Contravariant;
```
davidchambers: __figures/dependencies.png__ still shows "Contravariant Functor". ;)

README.md
```diff
@@ -28,7 +29,7 @@ structures:
 * [Bifunctor](#bifunctor)
 * [Profunctor](#profunctor)
 
-<img src="figures/dependencies.png" width="888" height="340" />
+<img src="figures/dependencies.png" width="1068" height="347" />
```
davidchambers: Rather than the image's natural dimensions, I think we want the dimensions of the scaled image on GitHub:

```html
<img src="figures/dependencies.png" width="888" height="289" />
```

README.md
```diff
@@ -186,6 +187,31 @@ method takes one argument:
...
+
+    u.contramap(f)
+
+1. `f` must be a function,
+
+    1. If `f` is not a function, the behaviour of `contramap` is
+       unspecified.
+    2. `f` can return any value.
+
+2. `contramap` must return a value of the same Functor
```
Avaq: "a value of the same **Functor**"?

rjmk: Good catch! I imagine this is a hangover from the `Contravariant Functor` name

README.md
joneshf: Sorry, Didn't realize we were scaling things :blush:.

joneshf:
Looks like everything has been addressed here. I'm merging. Thanks everyone!

davidchambers:
@joneshf, would you like to release v3.1.0?

Issue #130, opened 2016-03-24
Apply ?= Applicative, Chain ?= Monad


ccorcos:
I'm a little confused here. 

Is there any Functor that doesn't have a `.of` function? Isn't that just a constructor?

```js
Just.prototype.of = function(x) {
  return new Just(x)
}
```

By this logic, an Apply is an Applicative and a Chain is a Monad.

Are there any good examples of data structures that are particularly only some of these categories, but not all of them?

puffnfresh:
Tuples

puffnfresh:
Key-value maps

ccorcos:
I'm sorry but I don't understand your comment. Monoids are entirely disconnected from Functors:

![](https://github.com/fantasyland/fantasy-land/raw/master/figures/dependencies.png)

And I don't follow what you're saying with key-value maps either. There needs to be some way of constructing them, right?

puffnfresh:
A tuple would need to look like:

```js
Tuple.of = function(x) {
  return new Tuple(Monoid.empty, x);
};
```

A map needs a key as well as a value. The key means it can not be an Applicative.

joneshf:
> There needs to be some way of constructing them, right?

You can still use other functions that a key value map provides. You aren't restricted to only using functions from this specification.

CrossEye:
If generic `Tuples` aren't clear enough, think of a `Pair(x)` type like this:

```js
// map:: (a -> b) -> Pair(x, a) -> Pair(x, b)
```

If `x` represents, say, three-character strings, then `map(square, Pair('foo', 7)); //=> Pair('foo', 49)` without issue, but there is no way to create an arbitrary `Pair.of(3)`.  Where would it get the string?  That's the fundamental problem.

ccorcos:
> Where would it get the string? That's the fundamental problem.

I see. So `.of` can take only a single parameter. And that parameter must be a basic type? Well what is a basic type? For example, what if you passed an array of two items in `Pair.of(['foo', 7])`? 

What I'm really struggling with is discriminative examples. I finally understand how all this stuff works in regards to Maybes and Arrays, but thats because they work for just about all of them. But what are some examples that discriminate between these categories well?

- whats a semigroup that isn't also a monoid?
- whats foldable that isn't also traversable?
- whats a functor that isn't also an apply?
- whats an apply that isn't also and applicative or a chain?
- whats a chain or applicative that isn't a monad?

I've been trying to find answers to these all damn week! Any ideas?

[...]

PR #152, opened 2016-09-03
Add ChainRec specification

### LINKED ISSUE #151: Add ChainRec type class
> [TailRec](https://github.com/purescript/purescript-tailrec) - A type class which captures stack-safe monadic tail recursion. 

It would be nice to have a common specification for Tail Recursive Monads in JS community, so that libraries like [Free](https://github.com/safareli/free) could use the interface to be stack safe.

in PS `MonadRec`type class looks like this:
```hs
class Monad m <= MonadRec m where
  tailRecM :: forall a b. (a -> m (Either a b)) -> a -> m b
```

So we could have something like this without stack overflow:
```js
Identity.tailRecM((n) => {
  if (n == 0) {
    return Identity(Either.Right("DONE"))
  }
  return Identity(Either.Left(n - 1))
})(20000)
```

one thing is that there are multiple `Either` implementations in js and this spec should depend on bare minimum from any `Either` type. from what I have found the minimum api from Either type is to have a `cata` method. but before that let's see an implementation of `tailRecM` of `Identity`:

```js
const tailRec = (f) => (a) => {
  let v = f(a)
  while (!isDone(v)) {
    v = f(getValue(v))
  }
  return getValue(v)
}

const runIdentity = (i) => i.x

Identity.tailRecM = (f) => (a) => Identity(tailRec((v) => runIdentity(f(v)))(a))
```
So we have seen usage of it and part of it's implementation. now what we have left is implement `isDone` and `getValue` so that they are as generic as possible.
```js
const isDone = (e) => e.cata({
  Right: () => true,
  Left: () => false,
})

const getValue = (e) => e.cata({
  Right: (v) => v,
  Left: (v) => v,
})
```

so as it looks any Either with `cata` would work so users shouldn't be tied to some particular Either implementation (as long as it has `cata`). 
To note this Either implementation would work with `tailRecM`.
```js
const Either = {
  Right: (v) => ({ cata: (c) => c.Right(v) }),
  Left: (v) => ({ cata: (c) => c.Left(v) }),
}
```

---

The hardest was to implement `tailRecM` for `Task/Future` but i have made it and after we agree on some interface I would  create PRs for some popular `Task/Future` implementations

safareli:
Type of `ChainRec` is:

```hs
class Chain m <= ChainRec m where
  chainRec :: ((a -> c) -> (b -> c) -> a -> m c) -> a -> m b
```

Fixes #151


joneshf:
This is great! And the other issue was awesome, though I didn't keep up.

I'm a little confused about that type signature. Where does the `b` come from?

rjmk:
`c` is probably `t a b` if that helps. So the `b` comes from somewhere inside `a -> m c`.

e.g.

```js
// f :: Number -> Either Number String
var f = n => n > 0 ? Left(n - 1) : Right("Gotta get low, low, low")
```

I guess it's kinda confusing that the spec is polymorphic in `c`, but the function itself won't be

README.md
```diff
@@ -305,6 +306,30 @@ method takes one argument:
 
 2. `chain` must return a value of the same Chain
 
+### ChainRec
+
+A value that implements the ChainRec specification must also implement the Chain specification.
+
+1. `T.chainRec((done, next, value) => g(done(value)), i) `is equivalent to `g(i)`
```
rpominov: There is something wrong with `g`. It's certainly takes different types in `g(done(value))` and `g(i)` invocations.

Maybe you meant  `T.chainRec((done, next, value) => T.of(done(g(value))), i)`?

rpominov:
I'm curious maybe we could make it instance method:

```diff
- T.chainRec(f, i)
+ T.of(i).chainRec(f)
```

?

rpominov:
> I'm a little confused about that type signature. Where does the b come from?

I think type should look like this:

```
class Chain m <= ChainRec m where
  chainRec :: ((a -> c) -> (a -> c) -> a -> m c) -> a -> m a
```

README.md
safareli: `g` takes any type it's like `T.of` but I did not used `T.of` as `Chain` does not requite the method `of`.

i.e. this are equivalent:
`Identity.chainRec((done, next, value) => Identity.of(done(value)), 1)` ≡ `Identity.of(1)`

rpominov: Ah, right. It makes sense if we use `of` as `g`. Sorry.

Still curious if we could add more laws. Are there any laws for it in PureScript? If not, should we made up our own? Maybe we should add dependency on Monad after all, so we could use `of` and `map` in laws?

I just worry what is the value of having a method in the spec if we don't have laws that strictly restrict what it can do?

safareli:
> I'm curious maybe we could make it instance method:

We should depend least powerful specification. Now we only require structure to implement `Chain` spec, if we make it instance method then we would need `T.of` i.e. structure should implement `Monad`.  For `Monad` it could be derived for example:
```
T.prototype.monadRec = function(f){
  return this.chain(v => T.chainRec(f, v))
}
```
---

```
// type of `lastPagePostLikes` is `Task Number`
const lastPagePostLikes = Task.chainRec(
  (
    // this function creates some special value from one argument
    // the argument is of same type as `i`, `Page` in this case, it does the recursion 
    next,
    // this function creates some special value from one argument
    // value wrapped in `done` will be in `Task`as result of `Task.chainRec`
    // the argument type is `Number` in this case
    done,
    // on first call it's the `i` and every other invocation 
    // should be of same type `Page` but in JS you can not do that :p
    page
  ) => (
    page.nextPageURI
        // if current page has url to next page load the next page
        // here we have type `Task Page`
      ? HTTP.fetchJSON(page.nextPageURI)
        // here type would be `Task c`
        // but the `c` would be some `Either` like structure
        // so it would be like `Either Page Number`
        .map(next)
        //if we don't have next page url then this is the last page and return post likes
        // type here is `Task Number`
      : Task.of(page.posts.map(getLikes))
        // as this is what we wanted to actually get as the result wrap it in `done`
        // here type would be `Task c`
        // but the `c` would be some `Either` like structure
        // so it would be like `Either Page Number`
        .map(done)
  ),
  // this is `i` initial value. it's type is `Page`
  { nextPageURI : 'posts/1' , posts:[...]}
)
```

So type of `chainRec` in this case is 
```(
  (Page -> Either Page Number) ->
  (Number -> Either Page Number) ->
  Page ->
  Task (Either Page Number)
) -> Page -> Task Number```

replace `Number` with `b` and `Page` with `a`
```(
  (a -> Either a b) ->
  (b -> Either a b) ->
  a ->
  Task (Either a b)
) -> a -> Task b```

as value returned from `next/done` is use only internally it's irrelevant for client. so replace `Either a b` with c
```(
  (a -> c) ->
  (b -> c) ->
  a ->
  Task c
) -> a -> Task b```

replace `Task` with `m` and we get 
```((a -> c) -> (b -> c) -> a -> m c ) -> a -> m b```

File: README.md (hunk 1 above)
Comment (safareli) [inline README.md]: @paf31 @garyb do you seen know of any other "laws" for it?

**Comment by safareli** [comment]
@SimonRichardson should I also add the `law` to laws folder?
what should the name be for it?

**Comment by SimonRichardson** [comment]
Same name to keep consistency, that would be great 👍

**Comment by safareli** [comment]
I'm asking name of the law, like for monad we have `leftIdentity` and `rightIdentity`, but I don't know name for this one, I kinda discovered it when I was writing spec

**Comment by SimonRichardson** [comment]
Is that not just identity?

File: README.md (hunk 1 above)
Comment (rpominov) [inline README.md]: Still something bugs me about type of `g`. It seems like for any law in the spec where we have a function like that we can say "for any `g` of certain type". For example:

> `u.map(x => f(g(x)))` is equivalent to `u.map(g).map(f)`

for any:
- `u :: f a`
- `g :: a -> b` 
- `f :: b -> c`

> `a.of(f).ap(a.of(x))` is equivalent to `a.of(f(x))`

for any:
- `x :: a`
- `f :: a -> b`

> `m.chain(f).chain(g)` is equivalent to `m.chain(x => f(x).chain(g))`

for any:
- `m :: m a`
- `f :: a -> m b`
- `g :: b -> m c`

But here we can't say something like that about `g` because we can't identify it's type. I can't quite articulate why this is a problem, but it feels wrong.

Comment (rpominov) [inline README.md]: Have an idea about laws. We can implement `chainRec` using `chain`, right? Except that implementation will not be stack safe. But in ideal world they must be identical, at leas that what we want from `chainRec`. So we could say exactly that in a law. It could look like this:

> ```js
> T.chainRec(f, i)
> ```
> is equivalent to
>
> ```js
> const g = x => {
>   const r = f(next => ({next}), done => ({done}), x);
>   return 'next' in r ? g(r.next) : T.of(r.done)
> }
> T.of(i).chain(g)
> ```

Not sure that I implemented it using `chain` correctly, but you get the idea.

Comment (paf31) [inline README.md]: That's right. The first law says these two should give equal results, and the second says `chainRec` should do it in no more stack (asymptotically) than used by `f`.

Comment (safareli) [inline README.md]: @rpominov type of `g` is ` a => T a`. maybe I have not clearly expressed that with words. it looks like  there is no word on `f/g`, for example in monad spec . should I remove the line about `g`? or what you think is best wording for that?

What you think on something like this?
> `g` returns value of type `T` which contains value of it's first argument type.

---

About the other laws we could state this

>`T.chainRec((next,done,v) => p(v) ? f(v).map(done) : g(v).map(next), i)` is equivalent to 
>```js
(function step(v) {
  return p(v) ? f(v) : g(v).chain(step)
}(i))
>```
>But `chainRec` **must use constant stack space**


And types would be:
```hs
p :: a -> bool
g :: a -> T b
f :: a -> T a
step :: a -> T b
```

And we could also include this in `laws` like this:

```js
const someNameHere = t => eq => p => f => g => x => {
    const a = t.chainRec((next,done,v) => p(v) ? f(v).map(done) : g(v).map(next), x);
    const b = (function step(v) { return p(v) ? f(v) : g(v).chain(step) }(x));
    return eq(a, b);
};
```
But this must be tested with small inputs as `step` can blow the stack :D

If we call first law `identity` then what the name should be for this one? /cc @SimonRichardson

SimonRichardson: Without putting much thought into it, equivalence looks like it fits the
bill. If anyone disagrees then chime in.

rpominov: > type of `g` is `a => T a`

@safareli You've convinced me again that the law is correct. I don't know why my mind keep going back and forth on this :)

But if we have "equivalence" law, the first one seems redundant.

rpominov:
I still think it's good idea to make it depend on Monad, and use instance method as proposed in https://github.com/fantasyland/fantasy-land/pull/152#issuecomment-244557918

It's true that  we should depend on least powerful specification. But in FL we should also try to use instance methods if possible. That's how FL works — it allows to write generic code based on instance methods.

And I don't think we loose much if we make it dependent on Monad. Can we imagine a type that cannot implement `of` but can implement `chainRec`?

README.md
safareli: `equivalence` seems good. will remove first one

safareli:
@rpominov
> Can we imagine a type that cannot implement of but can implement chainRec?

maybe `Pair` (#130).

> But in FL we should also try to use instance methods if possible.

It could be instance method as Applicative's `of` is.
```js
T.chainRec = ....
T.prototype.chainRec = T.chainRec
```

safareli:
I could move clean up commit in different PR if necessary

rpominov:
Maybe we don't need dependency on Monad for `chainRec` to be static method e.g., `[1, 'a'].chainRec(f)` — seems like we don't need `of`.

> T.prototype.chainRec = T.chainRec

Yea, we should at least do this for consistency with `of` and `empty`.

SimonRichardson:
👍

README.md
```diff
@@ -305,6 +306,37 @@ method takes one argument:
 
 2. `chain` must return a value of the same Chain
 
+### ChainRec
+
+A value that implements the ChainRec specification must also implement the Chain specification.
+
+1. ```js
```
scott-christopher: [minor] These should use single backticks (inline code block) for consistency.

README.md
```diff
@@ -305,6 +306,37 @@ method takes one argument:
...
+   m.chainRec((next, done, v) => p(v) ? d(v).map(done) : n(v).map(next), i)
+   ```
+   is equivalent to
+
+   ```js
+   (function step(v) { return p(v) ? d(v) : n(v).chain(step); }(i))
+   ```
+   But stack usage of `m.chainRec(f, i)` must be at most a constant multiple of the stack usage of `f` itself. (equivalence)
```
scott-christopher: Perhaps this is worth putting into its own bullet point.
```
2. Stack usage of `m.chainRec(f, i)` must be at most a constant multiple of the stack usage of `f` itself.
```

README.md
```diff
@@ -305,6 +306,37 @@ method takes one argument:
...
+
+#### `chainRec` static method
+
+A Type which has a ChainRec must provide an `chainRec` method on itself
+or its `constructor` object. The `chainRec` method takes two arguments:
+
+    a.chainRec(f, i)
```
scott-christopher: [minor] These need some triple backticks for consistency.

README.md
```diff
@@ -305,6 +306,37 @@ method takes one argument:
...
+    a.constructor.chainRec(f, i)
+
+1. `f` must be a function which returns a value
+   1. If `f` is not a function, the behaviour of `chainRec` is unspecified.
+   2. `f` takes three arguments `next`, `done`, `value`
+       `next` and `done` are functions which take one argument and return some value.
```
scott-christopher: [minor] This new line isn't represented in the rendered markdown.

Perhaps this could split the definitions of `next`, `done` and `value` into sub-bullets.

>
`f` takes three arguments:
1. `next` is a function which takes one argument and can return any value
2. `done` is a function which takes one argument and returns the same type as the return value of `next`
3. `value` is some value of the same type as the argument for `next`

README.md
```diff
@@ -305,6 +306,37 @@ method takes one argument:
...
+   3. `f` must return a value of the same ChainRec which contains a value returned from either `done` or `next`
+   4. The third argument to `f` on it's first invocation should be `i`
```
scott-christopher: I'm unsure whether this point needs to be stated.

SimonRichardson:
This will now need a type signature has the PR #147 has landed.

README.md
rpominov: Maybe instead of this we should state that:

- The third argument to `f` must be of same type as `i`
- The first argument of `next` must be of same type as `i`

Basically mirror type signature `((a -> c) -> (b -> c) -> a -> m c) -> a -> m b`

README.md
safareli: Triple brackets are used only for type signatures.

Issue #139, opened 2016-05-09
publish to Bower registry


davidchambers:
I'd like to use the exported method names in projects which target browsers (in addition to Node). In order to do so, we must first:

1.  add environment detection to __index.js__, and export as `window` property in browser context;
2.  add __bower.json__; and
3.  run `bower register fantasy-land git://github.com/fantasyland/fantasy-land.git`.

I'll submit pull requests for 1 and 2 if maintainers are on board. :)

SimonRichardson:
@davidchambers can you make a PR.

davidchambers:
@SimonRichardson, you can perform step 3 now. :)

SimonRichardson:
This is done now!

PR #153, opened 2016-09-06
browser support

davidchambers:
Step 1 of #139

Is `FantasyLand` the best property name? Users are free to write `var FL = window.FantasyLand;` in their own code if a shorter name is desired.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,16 +1,28 @@
-module.exports = {
-  equals: 'fantasy-land/equals',
-  concat: 'fantasy-land/concat',
-  empty: 'fantasy-land/empty',
-  map: 'fantasy-land/map',
-  ap: 'fantasy-land/ap',
-  of: 'fantasy-land/of',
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,16 +1,28 @@
...
+    extend: 'fantasy-land/extend',
+    extract: 'fantasy-land/extract',
+    bimap: 'fantasy-land/bimap',
+    promap: 'fantasy-land/promap'
+  };
+
+  if (typeof module === 'object' && typeof module.exports === 'object') {
+    module.exports = mapping;
+  } else {
+    self.FantasyLand = mapping;
```
SimonRichardson: Where does `self` come from?

davidchambers: See davidchambers/Base64.js#27. `self` works in more environments than `window` does.

SimonRichardson: TIL: that `self` exists.

PR #157, opened 2016-09-11
Unbreak things

isiahmeadows:
Now the tests actually pass. Apparently, namespacing things apparently was a real *breaking* change... :wink:

Issue #88, opened 2015-02-11
Why can `of` be defined on the constructor?


CrossEye:
It's quite possible that I'm simply missing something simple, but I'm disturbed by this sentence (emphasis added):

> A value which has an Applicative must provide an `of` method on itself **or** its `constructor` object.

(All the same points will apply to Monoid's `empty` method, but it's Applicatives which are worrying me now.)

For a specification which is usually so prescriptive, this is surprisingly lax.  But that's not the real problem.  It seems to me that this makes it tremendously more difficult to write code that works across all Applicative Types; in fact, it probably makes it impossible.

A few points:

  * First of all, the Applicative laws, and Monad's _left identity_ are written only in terms of the instance version of `of`.  Should the specification be taken to read, for instance, "The appropriate one of `m.of(a).chain(f)` or `m.constructor.of(a).chain(f)` is equivalent to `f(a)` (left identity)" or some more precise version of the same?

  * It makes unwarranted assumptions about how I create types.  In these days of `Object.create`, why should the specification assume that I use constructor functions to define my types?  It's quite possible to do without them, and it's growing ever more popular to work that way.

  * It is possibly incoherent.  Since nothing specifies _which_ of the two I must implement, I can choose to implement the same `of` on both.  This would probably not cause an issue.  But nothing in the specification would prevent me from creating a conforming `of` on the instances and an unrelated `of` on the constructor or vice versa:

```javascript
    Maybe.of = function(val) {return new Just(val);};
    Maybe.prototype.of = function(val) {throw "Unknown Preposition";};
```

Does that conform to the specification as written so long as `Maybe.of` upholds the required laws?

Does this one so long as `Maybe.prototype.of` upholds the laws?:

```javascript
    Maybe.of = function(val) {throw "Unknown Preposition";};
    Maybe.prototype.of = function(val) {return new Just(val);};
```

If so, can anyone suggest a way that I can generically apply `of` to algebraic types without knowing for each specific type which version is being used?

Or... am I just missing something simple again? 

This was all brought to mind by a [recent Ramda issue][ri].

  [ri]: https://github.com/ramda/ramda/pull/784

michaelficarra:
[...]

PR #158, opened 2016-09-11
Single source of truth for everything

isiahmeadows:
This stemmed from [this short discussion](https://github.com/fantasyland/fantasy-land/pull/146#issuecomment-244973256). I think it would be a good idea to have a single source of truth for everything: the object itself (instead of an optional `constructor` method as well).

Given other outstanding PRs and issues, I'd expect this would need to wait a bit even if accepted, because it messes with so much. Also, I refactored a few bits of code to not use classes unnecessarily.


isiahmeadows:
Also, note that this is based on #157, so it carries that commit. [Here's that commit only.](https://github.com/fantasyland/fantasy-land/pull/158/commits/181555e91b11acb2c25e48160693328dbb09bdaf)

isiahmeadows:
(By the way, I'm not that attached to this. If it gets rejected, I'm not that concerned.)

README.md
```diff
@@ -254,27 +252,25 @@ implement the Functor and Foldable specifications.
 1. `t(u.sequence(f.of))` is equivalent to `u.map(t).sequence(g.of)`
 for any `t` such that `t(a).map(f)` is equivalent to `t(a.map(f))` (naturality)
 
-2. `u.map(F.of).sequence(F.of)` is equivalent to `F.of(u)` for any Applicative `F` (identity)
+2. `u.map(f.of).sequence(f.of)` is equivalent to `f.of(u)` for any Applicative `f` (identity)
 
-3. `u.map(x => new Compose(x)).sequence(Compose.of)` is equivalent to
-   `new Compose(u.sequence(F.of).map(v => v.sequence(G.of)))` for `Compose` defined below and any Applicatives `F` and `G` (composition)
+3. `u.map(compose.of).sequence(compose.of)` is equivalent to
```
rpominov: Something wrong here. `new Compose` is intentionally not the same as `Compose.of` see https://github.com/fantasyland/fantasy-land/pull/141

rpominov:
If we have `of` method only on values how do we create a value in the first place? :)

rpominov:
Another ways to improve things in this area:

- Replace "or" with "and" in "on itself or its `constructor`".
- State that in order to be compatible with FL one need to use JS "classes" (either via `function` or `class`) so it would be clear where to look for static `of/empty/etc` — on the class; Require only static `of/empty/etc`.

isiahmeadows:
@rpominov 

> If we have `of` method only on values how do we create a value in the first place? :)

Note that the examples use plain objects for this reason. It's somewhat Lua-like.

> Another ways to improve things in this area:
> 1. Replace "or" with "and" in "on itself or its  constructor ".

I was aiming to remove the latter. (Check out the green parts. `grep . 'constructor'` should bring absolutely nothing.)

> 2. Require only static  of/empty/etc  methods.
>
> Maybe additionally to \#1 or \#2: state that in order to be compatible with FL one need to use JS "classes" (either via `function` or `class`) so it would be clear where to look for static `of/empty/etc` — on the class.


[See this comment by @davidchambers.](https://github.com/fantasyland/fantasy-land/pull/146#issuecomment-245007036) That's why I removed references to constructor methods entirely.

README.md
isiahmeadows: I'll fix that when I get home to `compose.create` (the intended replacement method).

rpominov:
> Note that the examples use plain objects for this reason. It's somewhat Lua-like.

I still don't understand how one would use such API. Say someone gave me a type that compatible with FL; how do I create a value of that type? Or maybe FL don't have to support this use case (not sure).

Btw why do we switch to plain objects and `__proto__`? We can remove static methods but keep "classes", right?

> See this comment by @davidchambers. That's why I removed references to constructor methods entirely.

That make sense. Seems like having it on prototype is essential.

Btw, this PR is related to #88 (seems like it wasn't mentioned before).

README.md
```diff
@@ -254,27 +252,25 @@ implement the Functor and Foldable specifications.
...
-Compose.prototype.ap = function(x) {
-  return new Compose(this.c.map(u => y => u.ap(y)).ap(x.c));
-};
-
-Compose.prototype.map = function(f) {
-  return new Compose(this.c.map(y => y.map(f)));
-};
+var compose = {
+  of: function(x) {
+    return {__proto__: compose, c};
```
SimonRichardson: Really?

isiahmeadows: Would you like a different method of setting the prototype? (Keep in mind, I'm not actually that deeply invested in this. I'm merely encouraging exploration.)

isiahmeadows:
@rpominov WRT #88, I think I saw that once, but that's not what drove me to write this. (It is related, though.)

README.md
davidchambers: Why not rely on the `constructor` property as we currently do? We could write:

```javascript
//  Identity :: a -> Identity a
var Identity = function Identity(x) {
  return {
    constructor: {'fantasy-land/of': Identity},
    'fantasy-land/chain': function(f) { return f(x); },
    'fantasy-land/map': function(f) { return Identity(f(x)); },
    // ...
    value: x
  };
};
```

The nice thing about this approach is it allows a prototype-based implementation:

```javascript
//  Identity :: a -> Identity a
var Identity = function Identity(x) {
  if (!(this instanceof Identity)) return new Identity(x);
  this.value = x;
};

//  Identity.of :: a -> Identity a
Identity['fantasy-land/of'] = Identity;

//  Identity#chain :: Identity a ~> (a -> Identity b) -> Identity b
Identity.prototype['fantasy-land/chain'] = function(f) { return f(this.value); };

//  Identity#map :: Identity a ~> (a -> b) -> Identity b
Identity.prototype['fantasy-land/map'] = function(f) { return Identity(f(this.value)); };

// ...
```

I don't see a reason to use `__proto__`.

rpominov: > constructor: {'fantasy-land/of': Identity}
> Identity['fantasy-land/of'] = Identity

But didn't we want to remove static method?

rpominov: Anyway if we want to remove static method, I think we should simply _remove static method_ from current implementations, without switching to simple objects or anything.

isiahmeadows:
Given the immediate negative feedback (and the admitted non-idiomatic nature of it), I'm going to close it now. I think this was fully explored.

isiahmeadows:
#157 should still be looked at (it's just a series of bug fixes)

README.md
```diff
@@ -138,11 +138,10 @@ the Semigroup specification.
 empty :: Monoid m => () -> m
 ```
 
-A value which has a Monoid must provide an `empty` method on itself or
-its `constructor` object. The `empty` method takes no arguments:
+A value which has a Monoid must provide an `empty` method. The `empty`
+method takes no arguments:
 
     m.empty()
-    m.constructor.empty()
```
davidchambers: I believe we all agree that *this* change is a good idea. I've opened #162 which makes this change only.

PR #159, opened 2016-09-12
bower: add bower.json

davidchambers:
Step 2 of #139

Note that this only exposes __index.js__ since the other JavaScript files depend on fantasy-combinators which is not in the Bower registry.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,37 @@
+{
+  "name": "fantasy-land",
+  "description": "Specification for interoperability of common algebraic structures in JavaScript",
+  "license": "MIT",
+  "repository": {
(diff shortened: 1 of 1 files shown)


SimonRichardson:
I'd be happy to put fantasy-combinators on to bower if we want to make that a requirement!?

davidchambers:
> I'd be happy to put fantasy-combinators on to bower if we want to make that a requirement!?

Sure. If you perform the three steps listed in #139 for fantasy-combinators, we can then revisit this pull request.

SimonRichardson:
I'll release this without `fantasy-combinators` for now, otherwise we'll be here for a while!

PR #160, opened 2016-09-12
update dependency graph

davidchambers:
I find the existing dependency graph very helpful, but it's quite out of date. I've created an up-to-date ASCII version of the dependency graph. This will be easier to update as new algebra are added.

I was pleased to discover that GitHub renders `a` elements inside `pre` elements, which means we can have each node link to its corresponding section, obviating the need for the list of section links.

See the [rendered graph][1]. Pretty cool, I think!

[1]: https://github.com/davidchambers/fantasy-land/tree/graph#readme


joneshf:
Fun fact: SVG can contain anchor tags. Another fun fact: the dot grammar provides a way to attach urls to pretty much anything. So, we can do the linking with the current image.

I'm pretty much a :-1: on ascii art, because maintaining ascii art is not fun. But I'm a :+1: on updating stuff.

What we should do, is make the current dot file easier to maintain. We should also make the generation of the svg straight-forward. I can look into it though so the burden isn't all on you.

SimonRichardson:
Ascii one looks pants on mobile, not sure if that's good or bad - I'll leave my unbiased remark here though.

rpominov:
To use SVG with links sounds like a nice option, it's tricky to include SVG image into GitHub's markdown though: http://stackoverflow.com/questions/13808020/include-an-svg-hosted-on-github-in-markdown

joneshf:
And look who already went down this road: http://stackoverflow.com/a/21521184/1549047 !!

davidchambers:
Apparently I'm the authority on displaying SVGs in GitHub readmes (thanks to [Orthogonal][1], no doubt). As stated in my [second answer][2] it is possible with the help of [RawGit][3].


[1]: https://github.com/davidchambers/orthogonal
[2]: http://stackoverflow.com/a/24751219/312785
[3]: http://rawgit.com/

davidchambers:
For an even fancier version, see [type-class hierarchy][1]. So much information in so few words!


[1]: https://github.com/sanctuary-js/sanctuary-type-classes#type-class-hierarchy

davidchambers:
I've updated the diagram to include the names of the methods associated with the type classes.

Although this pull request is not the *ultimate* solution, it's a clear improvement over what we have now. Up-to-date content trumps other factors. It's unfortunate that the diagram does not look good on some mobile devices, but the current one does not paint a complete picture on *any* device. There's surely a more maintainable SVG solution, but I imagine the ASCII diagram will be updated more often than the current one has been. :stuck_out_tongue_winking_eye:

I suggest we merge this pull request. If someone would like to submit a pull request for an updated SVG with hyperlinks—along with a script for updating the image—I'll be thrilled! Let's not wait for the perfect solution, though, as we may be waiting a long time.

joneshf:
Makes sense. And thanks for prodding this issue. I totally dropped the ball.

Updated svg with hyperlinks and a script for updating the image: https://github.com/fantasyland/fantasy-land/compare/master...joneshf:dependencies Should I PR?

davidchambers:
> Should I PR?

Yes, please do. The .dot file is beautifully declarative!

joneshf:
Done: https://github.com/fantasyland/fantasy-land/pull/192

davidchambers:
Superseded by #192

PR #161, opened 2016-09-12
Setup travis

rpominov:
Tested that it works in my fork https://travis-ci.org/rpominov/fantasy-land/builds

After merging someone with write access will need to go to https://travis-ci.org/ and just turn it on.


SimonRichardson:
I've activated the travis CI - should require a push i believe.

PR #162, opened 2016-09-13
specify location of ‘empty’ method

davidchambers:
Commit message:

> An instance method is more flexible than a static method. It facilitates
>
> ```javascript
> > Identity([1, 2, 3]).empty()
> Identity([])
> ```
>
> since `Identity#empty` can dispatch to the inner value's `empty` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,11 +138,10 @@ the Semigroup specification.
 empty :: Monoid m => () -> m
 ```
 
-A value which has a Monoid must provide an `empty` method on itself or
-its `constructor` object. The `empty` method takes no arguments:
+A value which has a Monoid must provide an `empty` method. The `empty`
+method takes no arguments:
 
     m.empty()
-    m.constructor.empty()
 
 1. `empty` must return a value of the same Monoid
 
--- a/id.js
+++ b/id.js
@@ -15,10 +15,9 @@ Id.prototype[fl.concat] = function(b) {
 };
 
 // Monoid (value must also be a Monoid)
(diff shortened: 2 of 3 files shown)


CrossEye:
> I find `empty` most useful when an instance is not available, so I would still want a way to obtain an `empty` value statically.

That's not where I started, but it's where I ended up in the long discussion in #88.  I'm afraid this is one of those square peg/round hole situations that are bound to come up when you try to merge static and dynamic typing.  I don't know that there is a good solution.

davidchambers:
> I'd probably go a step further and prefer the `empty` instance to be a plain property value rather than a nullary method.

This was proposed in #82. This would mean that the Identity type could not satisfy Monoid, I believe. Does this bother you, @scott-christopher?

> The typical implementation of `Const.of` requires an `empty` value of the inner type, which makes it difficult to wire up if an instance is required.
>
> ```javascript
> Const.of = (_) => Const(OtherType.empty())
> ```

Hmm, I see the problem.

It seems we have four options:

1.  Require an instance method only, making the Const type sad. :cry:
2.  Require a static method only, making the Identity type sad. :cry:
3.  Continue to support both, but state that the instance method should be used if both are present.
4.  Do nothing.

Does anyone object to the third option? It's an improvement on the status quo, and shouldn't make anyone sad.

SimonRichardson:
3 sounds reasonable

davidchambers:
I'll open a new pull request shortly.

PR #163, opened 2016-09-13
favour m.empty over m.constructor.empty

davidchambers:
Supersedes #162

### DIFF
--- a/id.js
+++ b/id.js
@@ -6,7 +6,7 @@ function Id(a) {
 
 // Setoid
 Id.prototype[fl.equals] = function(b) {
-    return typeof this.value[fl.equals] === "function" ? this.value[fl.equals](b.value) : this.value === b.value;
+    return typeof this.value[fl.equals] === 'function' ? this.value[fl.equals](b.value) : this.value === b.value;
 };
 
 // Semigroup (value must also be a Semigroup)
@@ -15,10 +15,9 @@ Id.prototype[fl.concat] = function(b) {
 };
 
 // Monoid (value must also be a Monoid)
-Id[fl.empty] = function() {
-    return new Id(this.value[fl.empty] ? this.value[fl.empty]() : this.value.constructor[fl.empty]());
+Id.prototype[fl.empty] = function() {
+    return new Id(typeof this.value[fl.empty] === 'function' ? this.value[fl.empty]() : this.value.constructor[fl.empty]());
 };
-Id.prototype[fl.empty] = Id[fl.empty];
 
 // Foldable
 Id.prototype[fl.reduce] = function(f, acc) {
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ its `constructor` object. The `empty` method takes no arguments:
 
 1. `empty` must return a value of the same Monoid
 
+Implementations must favour `m.empty` over `m.constructor.empty`.
+
 ### Functor
 
 1. `u.map(a => a)` is equivalent to `u` (identity)


rpominov:
Would be great to also replace "or" with "and" in "method on itself or its constructor object" everywhere.

davidchambers:
> Would be great to also replace "or" with "and" in "method on itself or its constructor object" everywhere.

I would love to do this, but I don't think it's valid. It's possible to implement `Identity#empty` (an instance method) but not `Identity.empty` (a static method). What would be the result of evaluating `Identity.empty()`? There's no sensible answer. One could define an instance method, though:

```javascript
//  Identity#empty :: Identity a ~> () -> Identity a
Identity.prototype['fantasy-land/empty'] = function() {
  return Identity(this.value.empty());
};
```

If we're to require both instance method and static method, the Identity type (and many other types) will no longer be able to satisfy Monoid. We may decide that this is acceptable. If we do, though, we can get rid of the instance method as it will no longer serve a role not served by the static method (or static value).

rpominov:
Maybe FL shouldn't support use-case of delegating `empty` to inner type for the reason that it cannot do it generally anyway. `Identity` is just lucky to always have access to wrapped value. Types that do something useful usually don't have that access (e.g. Maybe, List, Task).

rpominov:
Another though: if we agree that [there are cases when we need static `empty`](https://github.com/fantasyland/fantasy-land/pull/162#issuecomment-246651027), the an `Identity` that doesn't have static `empty` won't work in those cases.

davidchambers:
> Maybe FL shouldn't support use-case of delegating `empty` to inner type for the reason that it cannot do it generally anyway.

This sounds reasonable to me.

One *can* define `empty` for the Maybe type (by considering `Nothing` the empty value), but it's not necessarily the preferred definition.

I'll submit a competing pull request to remove support for the instance method. :)

rpominov:
Maybe we should still keep both methods, but require that both versions must be present. There still can be cases where instance method is handy (if we always have a value and need to create the `empty` value).

Also I feel a bit like I'm blocking any solution in this area by proposing an alternative solution. So please ignore me and do whatever everybody feels right, if it only me proposing something else at some point :)

davidchambers:
> There still can be cases where instance method is handy (if we always have a value and need to create the `empty` value).

Having to write `.constructor` every now and then is a small price to pay for simplifying the spec and making life easier for type creators and library authors. Typing `m.constructor.empty` isn't onerous. :)

rpominov:
> Having to write .constructor

Ah, right. Forgot that even static method still can be accessed on a value via `constructor` 👍

davidchambers:
> I'll submit a competing pull request to remove support for the instance method. :)

#164

id.js
```diff
@@ -15,10 +15,9 @@ Id.prototype[fl.concat] = function(b) {
 };
 
 // Monoid (value must also be a Monoid)
-Id[fl.empty] = function() {
-    return new Id(this.value[fl.empty] ? this.value[fl.empty]() : this.value.constructor[fl.empty]());
+Id.prototype[fl.empty] = function() {
+    return new Id(typeof this.value[fl.empty] === 'function' ? this.value[fl.empty]() : this.value.constructor[fl.empty]());
```
safareli: It's sort of strange to have something in value['fantasy-land/empty'] which could not be called, I think it's better to not check and fail fast, so users will spot errors faster.

davidchambers: I agree, but I'm not sure the spec allows it:

> A value which has a Monoid must provide an `empty` method on itself or its `constructor` object.

`{'fantasy-land/empty': null, constructor: {'fantasy-land/empty': () => …}}` is valid, strictly speaking. Nowhere do we state that the method must be defined in *both* places, nor do we state that if a Fantasy Land property name is present its value must accord with the spec. I'd love to remove one or both of these sources of ambiguity.

davidchambers:
Closing in favour of #180

PR #168, opened 2016-09-18
Move unnecessary normal dependencies to devDependencies

rpominov:
These dependencies are not needed for `fantasy-land` npm package to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,10 @@
   "issues": {
     "url": "https://github.com/fantasyland/fantasy-land/issues"
   },
-  "dependencies": {
+  "devDependencies": {
+    "nodeunit": "0.9.x",
     "daggy": "0.0.x",
     "fantasy-combinators": "0.0.x"
-  },
-  "devDependencies": {
-    "nodeunit": "0.9.x"
   },
   "repository": {
     "type": "git",


davidchambers:
> These dependencies are not needed for `fantasy-land` npm package to work.

That depends on which parts of the package you're using. `require('fantasy-land')` works without `daggy` and `fantasy-combinators`, but `require('fantasy-land/laws/traversable')` requires these dependencies to be installed.

Are you suggesting the laws should be defined in a separate npm package, or that since they're only used for testing they should only not be installed in production?

rpominov:
Ah, good point. I forgot that package also exposes laws tests. Not sure what a right course of action here then. I'll close this PR for now, maybe we'll figure out something later.

safareli:
Use of duggy in `laws/traversable` is sort of redundant in think, we could replace that lines with this:
```diff
-const {tagged} = require('daggy');
-const Compose = tagged('c');
+function Compose(c) {
+  if (!(this instanceof Compose)) {
+    return new Compose(c)
+  }
+  this.c = c
+}
```
(actually same is in `id_test` too: `const Sum = tagged('v');`)

And about fantasy-combinators we use `Identity` (12 times), `apply` (3 times) and `compose` (3 times). there are 9 functions in fantasy-combinators and we use 3 and each is just one line, so maybe it's better to could just create `combinators.js` and include this 3 functions in there?

SimonRichardson:
I'm not a fan of code duplication for the sake of it :|

safareli:
On one side a) we have 2 dependencies and on other side b) duplication of 3 online functions and expansion of 2 constructors.

If we go with b) FL will have 0 dependency so library [contributors](https://github.com/Avaq/Fluture/pull/26#discussion_r79305928) [will](https://github.com/sanctuary-js/sanctuary-type-classes/pull/2#issuecomment-247303007) be easily convinced to use FL as dependency. of course this point has value in case we want libs to depend on FL, which I think is what we should be aiming for as it will make it clear what version of FL a lib conforms to without digging into it's docs.

rpominov:
Good point about determining to which version of FL a library conforms to. Didn't thought about it this way.

safareli:
Also it's handy to see libs which depend on FL in [npm](https://www.npmjs.com/browse/depended/fantasy-land)

rpominov:
Wow, that's a good alternative to [implementations.md](https://github.com/fantasyland/fantasy-land/blob/master/implementations.md) that can be more complete and automatically up to date!

SimonRichardson:
:sigh: not invented here syndrome.

OUT

SimonRichardson:
I think if it boils down to this it's beyond pathetic. I just don't see value in it. If we think this, then I'd go the whole hog and remove all the other libraries in fantasy-land and just host a markdown file.

PR #173, opened 2016-09-21
laws/functor: fix composition

### LINKED ISSUE #172: cannot use laws/functor to test composition property
[`identityʹ`][1]:

> ```javascript
> const identityʹ = t => eq => x => {
>     const a = t(x)[map](identity);
>     const b = t(x);
>     return eq(a, b);
> };
> ```

[`composition`][2]:

> ```javascript
> const composition = t => eq => x => {
>     const a = t(x)[map](compose(identity)(identity));
>     const b = t(x)[map](identity)[map](identity);
>     return eq(a, b);
> };
> ```

Consider `a` in `composition`:

    a = t(x)[map](compose(identity)(identity))

Assume a proof for `compose(identity)(identity) = identity`:

    a = t(x)[map](identity)

The identity property tells us that `t(x)[map](identity) = t(x)`, so:

    a = t(x)

Consider `b` in `composition`:

    b = t(x)[map](identity)[map](identity)

The identity property tells us that `t(x)[map](identity) = t(x)`, so:

    b = t(x)[map](identity)
    b = t(x)

I think all we're actually proving is `compose(identity)(identity) = identity`, which isn't at all what we're hoping to prove. :stuck_out_tongue_winking_eye:

For `composition` to be useful I think it needs to take `f :: b -> c` and `g :: a -> b` as arguments.

[1]: https://github.com/fantasyland/fantasy-land/blob/1.0.1/laws/functor.js#L15-L19
[2]: https://github.com/fantasyland/fantasy-land/blob/1.0.1/laws/functor.js#L21-L25

davidchambers:
Fixes #172

### DIFF
--- a/laws/functor.js
+++ b/laws/functor.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const {identity, compose} = require('fantasy-combinators');
 const {map} = require('..');
 
 /*
@@ -12,18 +11,18 @@ const {map} = require('..');
 
 */
 
-const identityʹ = t => eq => x => {
-    const a = t(x)[map](identity);
-    const b = t(x);
+const identity = of => eq => x => {
+    const a = of(x)[map](x => x);
+    const b = of(x);
     return eq(a, b);
 };
 
-const composition = t => eq => x => {
-    const a = t(x)[map](compose(identity)(identity));
-    const b = t(x)[map](identity)[map](identity);
+const composition = of => eq => x => f => g => {
+    const a = of(x)[map](x => f(g(x)));
+    const b = of(x)[map](g)[map](f);
     return eq(a, b);
 };
 
-module.exports = { identity: identityʹ
-                 , composition 
+module.exports = { identity
+                 , composition
                  };


laws/functor.js
```diff
@@ -12,18 +11,18 @@ const {map} = require('..');
...
+    const a = of(x)[map](x => x);
+    const b = of(x);
     return eq(a, b);
 };
 
-const composition = t => eq => x => {
-    const a = t(x)[map](compose(identity)(identity));
-    const b = t(x)[map](identity)[map](identity);
+const composition = of => eq => x => f => g => {
+    const a = of(x)[map](x => f(g(x)));
```
SimonRichardson: why not `of(x)[map](compose)` - call a spade a spade?

davidchambers: This more closely matches what appears in the spec. Also, the `fantasy-combinators` dependency seems unnecessary to me, since `x => f(g(x))` and the other expressions it allows us to write without points are not verbose to begin with.

PR #180, opened 2016-10-03
require static methods to be defined on type representatives

### LINKED ISSUE #176: require static methods to be defined on type representatives
_I spent a long time writing a lot of words, but I've decided to delete many of them and start with an example. Hopefully this makes the problem clear._

---

Let's use [JSVerify][1] to define a generator for possibly empty values of any type which satisfies both Monoid and Applicative:

```javascript
//    ApplicativeMonoidArb :: (Applicative m, Monoid (m a)) =>
//                              { fantasy-land/empty :: () -> m a } ->
//                              { fantasy-land/of :: a -> m a } ->
//                              Arbitrary a ->
//                              Arbitrary (m a)
const ApplicativeMonoidArb = $empty => $of => arb =>
  jsc.oneof(jsc.constant($empty['fantasy-land/empty']()),
            arb.smap($of['fantasy-land/of']));
```

Usage example:

```javascript
//    MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
const MaybeArb = ApplicativeMonoidArb(S.Maybe)(S.Maybe);
```

`MaybeArb` could then be used to generate values for property-based testing:

```javascript
jsc.assert(jsc.forall(MaybeArb(jsc.string), m => {
  //  Assert that ‘m’ satisfies various laws...
}));
```

Why are `{ fantasy-land/empty :: () -> m a }` and `{ fantasy-land/of :: a -> m a }` separate? (It forces us to provide `S.Maybe` twice, which is tedious.) The reason is that the spec allows this:

    { constructor :: { fantasy-land/empty :: () -> Maybe a }
    , fantasy-land/of :: Maybe a ~> b -> Maybe b
    , ...
    }

As a result there's no guarantee that there exists a "record" with both `fantasy-land/empty` and `fantasy-land/of` fields. :(

Defining `MaybeArb` was straightforward, but let's also consider the case in which we're given a value `x` of type `(Applicative m, Monoid (m a)) => m a` from which we'd like to generate other values. How would we use `ApplicativeMonoidArb` in this case? It's not pretty:

```javascript
//    XArb :: Arbitrary a -> Arbitrary (X a)
const XArb = ApplicativeMonoidArb(typeof x['fantasy-land/empty'] === 'function' ? x : x.constructor)(typeof x['fantasy-land/of'] === 'function' ? x : x.constructor);
```

This is our current model:

    +------------------+                     +------------------+
    |                  |                     |                  |
    |      t :: T      | --> constructor --> |        ??        |
    |                  |                     |                  |
    +------------------+                     +------------------+
       |                                        |
       +--> fantasy-land/empty .... AND/OR .... +--> fantasy-land/empty
       |                                        |
       +--> fantasy-land/concat                 |
       |                                        |
       +--> fantasy-land/map                    |
       |                                        |
       +--> fantasy-land/ap                     |
       |                                        |
       +--> fantasy-land/of ....... AND/OR .... +--> fantasy-land/of

I propose that we adopt this model:

    +------------------+                     +------------------+
    |                  |                     |                  |
    |      t :: T      | --> constructor --> |  T :: TypeRep T  |
    |                  |                     |                  |
    +------------------+                     +------------------+
       |                                        |
       +--> fantasy-land/concat                 +--> fantasy-land/empty
       |                                        |
       +--> fantasy-land/map                    +--> fantasy-land/of
       |
       +--> fantasy-land/ap

How does this differ from the current model?

  - Everything is defined in exactly one place.

  - Static methods are clearly separated from instance methods.

  - Rather than `??` which may or may not have some useful properties we have a [type representative][2] with known properties.

How would our example look in this world?

```javascript
//    ApplicativeMonoidArb :: (Applicative m, Monoid (m a)) => TypeRep m -> Arbitrary a -> Arbitrary (m a)
const ApplicativeMonoidArb = typeRep => arb =>
  jsc.oneof(jsc.constant(typeRep['fantasy-land/empty']()),
            arb.smap(typeRep['fantasy-land/of']));
```

This type signature is much clearer!

Updated usage example (given a type representative):

```javascript
//    MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
const MaybeArb = ApplicativeMonoidArb(S.Maybe);
```

We need only provide `S.Maybe` once now. :)

Updated usage example (given a value):

```javascript
//    XArb :: Arbitrary a -> Arbitrary (X a)
const XArb = ApplicativeMonoidArb(x.constructor);
```

Where did all our ternaries go? ;)

---

Now that we have prefixed method names and `ap` is consistent with `map` and `chain`, I believe the `{empty,of,chainRec}` ["schizophrenia"][3] to be the most important unresolved issue. It has already been discussed a great deal: #16, #88, #97, #158, #162, #163, #164, sanctuary-js/sanctuary-type-classes#3. It would be wonderful to agree on a solution (which may or may not be the one proposed above) and release an update to the spec.

[1]: https://github.com/jsverify/jsverify
[2]: https://sanctuary.js.org/#type-representatives
[3]: https://github.com/fantasyland/fantasy-land/issues/88#issuecomment-74401378

davidchambers:
Closes #176

> ## Type representatives
>
> Certain behaviours are defined from the perspective of a member of a type. Other behaviours do not require a member. Thus certain algebras require a type to provide a value-level representative (with certain properties). The Identity type, for example, could provide `Id` as its type representative: `Id :: TypeRep Identity`.
>
> If a type provides a type representative, each member of the type must have a `constructor` property which is a reference to the type representative.

I imagine there are several things I neglected to update. Let me know if you think of one. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,17 +48,14 @@ have dependencies on other algebras which must be implemented.
 
 ## Prefixed method names
 
-In order to add compatibility with Fantasy Land to your library,
-you need to add methods that you want to support with `fantasy-land/` prefix.
-For example if a type implements Functors' [`map`][], you need to add `fantasy-land/map` method to it.
-The code may look something like this:
+In order for a data type to be compatible with Fantasy Land, its values must
+have certain properties. These properties are all prefixed by `fantasy-land/`.
+For example:
 
 ```js
-MyType.prototype['fantasy-land/map'] = MyType.prototype.map
+//  MyType#fantasy-land/map :: MyType a ~> (a -> b) -> MyType b
+MyType.prototype['fantasy-land/map'] = ...
 ```
-
-It's not required to add unprefixed methods (e.g. `MyType.prototype.map`)
-for compatibility with Fantasy Land, but you're free to do so of course.
 
 Further in this document unprefixed names are used just to reduce noise.
 
@@ -69,12 +66,23 @@ var fl = require('fantasy-land')
 
 // ...
 
-MyType.prototype[fl.map] = MyType.prototype.map
+MyType.prototype[fl.map] = ...
 
 // ...
 
 var foo = bar[fl.map](x => x + 1)
 ```
+
+## Type representatives
+
+Certain behaviours are defined from the perspective of a member of a type.
+Other behaviours do not require a member. Thus certain algebras require a
+type to provide a value-level representative (with certain properties). The
+Identity type, for example, could provide `Id` as its type representative:
+`Id :: TypeRep Identity`.
+
+If a type provides a type representative, each member of the type must have
+a `constructor` property which is a reference to the type representative.
 
 ## Algebras
 
@@ -129,8 +137,8 @@ A value which has a Semigroup must provide a `concat` method. The
 A value that implements the Monoid specification must also implement
 the [Semigroup](#semigroup) specification.
 
-1. `m.concat(m.empty())` is equivalent to `m` (right identity)
-2. `m.empty().concat(m)` is equivalent to `m` (left identity)
+1. `m.concat(m.constructor.empty())` is equivalent to `m` (right identity)
+2. `m.constructor.empty().concat(m)` is equivalent to `m` (left identity)
 
 #### `empty` method
 
@@ -138,10 +146,14 @@ the [Semigroup](#semigroup) specification.
 empty :: Monoid m => () -> m
 ```
 
-A value which has a Monoid must provide an `empty` method on itself or
-its `constructor` object. The `empty` method takes no arguments:
+A value which has a Monoid must provide an `empty` function on its
+[type representative](#type-representatives):
 
-    m.empty()
+    M.empty()
+
+Given a value `m`, one can access its type representative via the
+`constructor` property:
+
     m.constructor.empty()
 
 1. `empty` must return a value of the same Monoid
@@ -206,9 +218,9 @@ method takes one argument:
 A value that implements the Applicative specification must also
 implement the [Apply](#apply) specification.
 
-1. `v.ap(a.of(x => x))` is equivalent to `v` (identity)
-2. `a.of(x).ap(a.of(f))` is equivalent to `a.of(f(x))` (homomorphism)
-3. `a.of(y).ap(u)` is equivalent to `u.ap(a.of(f => f(y)))` (interchange)
+1. `v.ap(A.of(x => x))` is equivalent to `v` (identity)
+2. `A.of(x).ap(A.of(f))` is equivalent to `A.of(f(x))` (homomorphism)
+3. `A.of(y).ap(u)` is equivalent to `u.ap(A.of(f => f(y)))` (interchange)
 
 #### `of` method
 
@@ -216,15 +228,20 @@ implement the [Apply](#apply) specification.
 of :: Applicative f => a -> f a
 ```
 
-A value which has an Applicative must provide an `of` method on itself
-or its `constructor` object. The `of` method takes one argument:
+A value which has an Applicative must provide an `of` function on its
+[type representative](#type-representatives). The `of` function takes
+one argument:
 
-    a.of(b)
-    a.constructor.of(b)
+    F.of(a)
+
+Given a value `f`, one can access its type representative via the
+`constructor` property:
+
+    f.constructor.of(a)
 
 1. `of` must provide a value of the same Applicative
 
-    1. No parts of `b` should be checked
+    1. No parts of `a` should be checked
 
 ### Foldable
 
@@ -333,10 +350,10 @@ method takes one argument:
 
 A value that implements the ChainRec specification must also implement the [Chain](#chain) specification.
 
-1. `m.chainRec((next, done, v) => p(v) ? d(v).map(done) : n(v).map(next), i)`
+1. `M.chainRec((next, done, v) => p(v) ? d(v).map(done) : n(v).map(next), i)`
    is equivalent to
    `(function step(v) { return p(v) ? d(v) : n(v).chain(step); }(i))` (equivalence)
-2. Stack usage of `m.chainRec(f, i)` must be at most a constant multiple of the stack usage of `f` itself.
+2. Stack usage of `M.chainRec(f, i)` must be at most a constant multiple of the stack usage of `f` itself.
 
 #### `chainRec` method
 
@@ -344,11 +361,16 @@ A value that implements the ChainRec specification must also implement the [Chai
 chainRec :: ChainRec m => ((a -> c, b -> c, a) -> m c, a) -> m b
 ```
 
-A Type which has a ChainRec must provide a `chainRec` method on itself
-or its `constructor` object. The `chainRec` method takes two arguments:
+A Type which has a ChainRec must provide a `chainRec` function on its
+[type representative](#type-representatives). The `chainRec` function
+takes two arguments:
 
-    a.chainRec(f, i)
-    a.constructor.chainRec(f, i)
+    M.chainRec(f, i)
+
+Given a value `m`, one can access its type representative via the
+`constructor` property:
+
+    m.constructor.chainRec(f, i)
 
 1. `f` must be a function which returns a value
     1. If `f` is not a function, the behaviour of `chainRec` is unspecified.
@@ -566,7 +588,7 @@ be equivalent to that of the derivation (or derivations).
 2. It's discouraged to overload the specified methods. It can easily
    result in broken and buggy behaviour.
 3. It is recommended to throw an exception on unspecified behaviour.
-4. An `Id` container which implements all methods is provided in
+4. An `Id` container which implements many of the methods is provided in
    `id.js`.
 
 
--- a/id_test.js
+++ b/id_test.js
@@ -23,8 +23,6 @@ const Id = require('./id');
 const Sum = tagged('v');
 Sum[of] = Sum;
 Sum[empty] = () => Sum('');
-Sum.prototype[of] = Sum[of];
-Sum.prototype[empty] = Sum[empty];
 Sum.prototype[map] = function(f) {
   return Sum(f(this.v));
 };
@@ -90,8 +88,8 @@ exports.monad = {
 };
 
 exports.monoid = {
-  leftIdentity: test(x => monoid.leftIdentity(Id[of](Sum[empty]()))(equality)(Sum[of](x))),
-  rightIdentity: test(x => monoid.rightIdentity(Id[of](Sum[empty]()))(equality)(Sum[of](x))),
+  leftIdentity: test(x => monoid.leftIdentity(Sum)(equality)(x)),
+  rightIdentity: test(x => monoid.rightIdentity(Sum)(equality)(x)),
 };
 
 // Semigroup tests are broken otherwise for this.
--- a/id.js
+++ b/id.js
@@ -16,11 +16,7 @@ Id.prototype[fl.concat] = function(b) {
   return new Id(this.value[fl.concat](b.value));
 };
 
-// Monoid (value must also be a Monoid)
-Id[fl.empty] = function() {
-  return new Id(this.value[fl.empty] ? this.value[fl.empty]() : this.value.constructor[fl.empty]());
-};
-Id.prototype[fl.empty] = Id[fl.empty];
+// Monoid is not satisfiable since the type lacks a universal empty value
 
 // Foldable
 Id.prototype[fl.reduce] = function(f, acc) {
@@ -62,7 +58,6 @@ Id[fl.chainRec] = function(f, i) {
   }
   return new Id(state.value);
 };
-Id.prototype[fl.chainRec] = Id[fl.chainRec];
 
 // Extend
 Id.prototype[fl.extend] = function(f) {
@@ -73,7 +68,6 @@ Id.prototype[fl.extend] = function(f) {
 Id[fl.of] = function(a) {
   return new Id(a);
 };
-Id.prototype[fl.of] = Id[fl.of];
 
 // Comonad
 Id.prototype[fl.extract] = function() {
--- a/laws/applicative.js
+++ b/laws/applicative.js
@@ -7,29 +7,29 @@ const {of, ap} = require('..');
 
 ### Applicative
 
-1. `v.ap(a.of(x => x))` is equivalent to `v` (identity)
-2. `a.of(x).ap(a.of(f))` is equivalent to `a.of(f(x))` (homomorphism)
-3. `a.of(y).ap(u)` is equivalent to `u.ap(a.of(f => f(y)))` (interchange)
+1. `v.ap(A.of(x => x))` is equivalent to `v` (identity)
+2. `A.of(x).ap(A.of(f))` is equivalent to `A.of(f(x))` (homomorphism)
+3. `A.of(y).ap(u)` is equivalent to `u.ap(A.of(f => f(y)))` (interchange)
 
 **/
 
-const identityʹ = t => eq => x => {
-  const a = t[of](x)[ap](t[of](identity));
-  const b = t[of](x);
+const identityʹ = T => eq => x => {
+  const a = T[of](x)[ap](T[of](identity));
+  const b = T[of](x);
   return eq(a, b);
 };
 
-const homomorphism = t => eq => x => {
-  const a = t[of](x)[ap](t[of](identity));
-  const b = t[of](identity(x));
+const homomorphism = T => eq => x => {
+  const a = T[of](x)[ap](T[of](identity));
+  const b = T[of](identity(x));
   return eq(a, b);
 };
 
-const interchange = t => eq => x => {
-  const u = t[of](identity);
+const interchange = T => eq => x => {
+  const u = T[of](identity);
 
-  const a = t[of](x)[ap](u);
-  const b = u[ap](t[of](thrush(x)));
+  const a = T[of](x)[ap](u);
+  const b = u[ap](T[of](thrush(x)));
   return eq(a, b);
 };
 
--- a/laws/chainrec.js
+++ b/laws/chainrec.js
@@ -6,14 +6,14 @@ const {chain, map, chainRec} = require('..');
 
 ### ChainRec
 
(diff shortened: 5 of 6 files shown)


README.md
```diff
@@ -129,22 +140,28 @@ A value which has a Semigroup must provide a `concat` method. The
...
 
-1. `m.concat(m.empty())` is equivalent to `m` (right identity)
-2. `m.empty().concat(m)` is equivalent to `m` (left identity)
+1. `m.concat(m.constructor.empty)` is equivalent to `m` (right identity)
+2. `m.constructor.empty.concat(m)` is equivalent to `m` (left identity)
 
-#### `empty` method
+<div id="empty-method"></div>
+
+#### `empty` value
```
davidchambers: The empty `div` allows existing links to this section to continue to work.

davidchambers:
The term "type representative" was suggested by @tel in https://github.com/sanctuary-js/sanctuary/pull/64#issuecomment-123129673. [Type representatives][1] are also described in the Sanctuary documentation.


[1]: https://sanctuary.js.org/#type-representatives

rpominov:
I like a lot this idea. Still wonder maybe we could make this not a breaking change or at least less breaking? The piece that concerns me is transition from `empty` function to `empty` value. What if we hold this particular change until the next major release? This way it will be less burden for implementers, their types most likely to just continue to be compatible, if we don't do the `empty() -> empty` part.

README.md
```diff
@@ -76,6 +76,17 @@ MyType.prototype[fl.map] = MyType.prototype.map
 var foo = bar[fl.map](x => x + 1)
 ```
 
+## Type representatives
+
+Certain behaviours are defined from the perspective of a member of a type.
+Other behaviours do not require a member. Thus certain algebras require a
+type to provide a value-level representative (with certain properties). The
+Identity type, for example, could provide `Id` as its type representative:
+`Id :: TypeRep Identity`.
```
davidchambers: In Haskell types are "real" in the sense that one actually references types in one's programs:

```haskell
inc :: Int -> Int
inc x = x + 1
```

In JavaScript types are not referenced (except in comments):

```javascript
//    inc :: Number -> Number
const inc = x => x + 1;
```

In JavaScript the *idea* of types exists—as evidenced by the use of type signatures in comments—but there's no "type level"; only a "value level". Since we can't reference types at the type level, we must reference them at the value level. So, for each type we'd like a value which represents it. We call such values type representatives.

Let's consider some examples:

  - `Number` is the representative of the Number type. `Number` is the sole inhabitant of the `TypeRep Number` type.

  - `Id` is the representative of the Identity type. `Id` is the sole representative of the `TypeRep Identity` type.

The formatting above is very much intentional. The Number type exists only as an idea, so should not be formatted as code. `Number`, on the other hand, is a JavaScript value which represents the Number type (in addition to its role as a function from `Any` to `Number`).

It took me a while to come to terms with the idea of type representatives. This sort of thinking is not natural in a dynamically typed language. I suggest reading sanctuary-js/sanctuary#64 if you have not already done so.

safareli:
It makes sense with `chainRec` and `of` as they either exist for some Type or not and they don't need instance value to work properly. 

We have `Semigroup` which needs instances to exist and Identity can be `Semigroup` if value in it is also a `Semigroup` but what I don't like is that now we can't define `Monoid` instance for `Identity` which holds Monoid value. One way would be to have specific Identities for each type like this:
```js
const makeId = M => {
  function Identity(value){...}
  Identity.empty = Identity(M.empty)
  return Identity
}
```

Do you see any way to make `Monoid m => Identity m` a monoid?

rpominov:
> now we can't define Monoid instance for Identity which holds Monoid value.

Could we do it before though? Technically maybe, but I don't think it was very useful Monoid in practice.

davidchambers:
> Still wonder maybe we could make this not a breaking change or at least less breaking? The piece that concerns me is transition from `empty` function to `empty` value.

I believe I can provide irrefutable logical reasoning for the `empty` change. Let's find out. :smile:

First, let's consider the question of breaking versus less breaking. From a [SemVer][1] perspective there's no distinction: *any* breaking change requires a major version increment. Were we to stick with an `empty` method for now, would this pull request still be breaking? The answer is yes (emphasis mine):

> A value which has an Applicative must provide an `of` method on itself __or__ its `constructor` object.

Currently it's valid to define `of` only on members of a type. If this pull request is merged, types defined in this way will no longer satisfy the requirements of Applicative. The location changes for `empty` and `chainRec` are also breaking changes.

Even without the `empty` change, this pull request will necessitate a major version increment if merged.

We should also consider the burden changes will place on authors of ADT libraries. I imagine this is what you had in mind when you wrote "less breaking".

Aside from the `empty` change, the burden this pull request may place on authors of ADT libraries is small. In most cases *nothing* would be required (though obsolete methods should be deleted to avoid confusion).

The next question, then, is whether we greatly increase the burden on authors of ADT libraries by changing the type of `empty` from `Monoid m => () -> m` to `Monoid m => m`.

I believe the answer is no. Simply by *moving* the method from the members to the type representative we preclude certain types—Identity, for example—from satisfying the requirements of Monoid. For some ADT libraries this will require documentation updates and major version increments of their own. This may require a significant amount of work, but this work would be required whether or not `empty` is changed from a method to a value.

The question then becomes why it was decided that `empty` should be a method in the first place. The reason was that certain types can be monoidal when the "inner" type is monoidal. This allows `Id('abc').empty()` to evaluate to `Id('')`. Since `empty` will no longer be an instance method, though, this will no longer be possible. If this is no longer possible, one of the two arguments against #82 will no longer be valid. The other argument against #82 was that it would be a breaking change necessitating a major version increment. This argument no longer applies either as the other changes in this pull request already commit us to a major version increment.

I'd love to merge this pull request and publish `fantasy-land@2.0.0`. Upgrading from 1.x.x to 2.x.x should not be difficult, in most cases, and many projects will be able to go straight from 0.2.x to 2.x.x.


[1]: http://semver.org/

safareli:
@rpominov yes we could:
```js
Id.prototype[fl.empty] = function() {
    return new Id(typeof this.value[fl.empty] == 'function' ? this.value[fl.empty]() : this.value.constructor[fl.empty]());
};		
```
looks like it's not that useful don't know, and this technique would not work for those types which don't hold value in moment of calling empty (IO/Task/Function) (not sure on this)

README.md
safareli: The part which confused me was `Id :: TypeRep Identity`  (`Id` and `Identity` are different when `Number` and `Number` are same) is it because we assume definition of Identity would be something like this `type Identity a = Id a`?

So if we have:
```hs
type Maybe a = Just a | Nothing
```

Then is this true:?
```hs
Just a :: TypeRep (Maybe a)
Nothing :: TypeRep (Maybe a)
```

tel: In Haskell at least those last equations are wrong. 

- Just constructs values at the Maybe type, it's not a representative of the type. In particular, a type should be represented in the shape of the type, not its values.
- Maybe is an interesting example since it's a type constructor, a type function. If we want to represent this is a value-level type representative then `TypeRep Maybe` and `TypeRep X` must be able to be combined into `TypeRep (Maybe X)`. Perhaps the type reps of constructors are functions, or probably better (since it'll be easier to analyze) you could have something that gives you `($$) :: TypeRep f -> TypeRep a -> TypeRep (f a)`

tel: If that $$ is data then you can easily analyze it!

rpominov:
> I imagine this is what you had in mind when you wrote "less breaking".

Yea.

> I'd love to merge this pull request and publish fantasy-land@2.0.0. Upgrading from 1.x.x to 2.x.x should not be difficult, in most cases, and many projects will be able to go straight from 0.2.x to 2.x.x.

But on the second thought, I agree. This should be done eventually, and not to much of a burden in any case.

rpominov:
@safareli Yea, but seems like it can be done only for Identity type. And even though, Identity itself can be useful sometimes (as a noop type), that particular Monoid isn't useful at all.

README.md
davidchambers: > The part which confused me was `Id :: TypeRep Identity` (`Id` and `Identity` are different when `Number` and `Number` are same) is it because we assume definition of Identity would be something like this `type Identity a = Id a`?

When we talk about "the Identity type" we're referring to a concept with no manifestation in JavaScript. When we talk about "`Id`" we're referring to a real thing: an identifier we could evaluate in a REPL. `Id` lives in the world of values to the left of the colons; `Identity` lives in the world of types to the right of the colons.

Since the world of values and the world of types are separate, we needn't worry about naming collisions. This is valid:

```haskell
Identity :: a -> Identity a
```

`Identity` on the left of the `::` is a value, whereas `Identity` on the right is a type (a type constructor, actually).

Similarly, we could name our type representative `Identity` rather than `Id` without conflict:

```haskell
Identity :: TypeRep Identity
```

I hope this clarifies things a little. :)

davidchambers:
> > now we can't define `Monoid` instance for `Identity` which holds Monoid value.
>
> Could we do it before though? Technically maybe, but I don't think it was very useful Monoid in practice.

This pull request has several predecessors. The first one I submitted was #162, which required `empty` to be defined on members to enable Identity to satisfy the requirements of Monoid. In the discussion thread @scott-christopher made an insightful [comment][1] which began:

> I find `empty` most useful when an instance is not available, so I would still want a way to obtain an `empty` value statically.

If `empty` is to be defined in one place only (as I strongly believe it should be), we must choose between allowing Identity (and possibly other types) to satisfy the requirements of Monoid, and allowing Const (and possibly other types) to do so. I can't recall ever being led astray by Scott's suggestions—he has a much better grasp of this stuff than I do—so I'm inclined to believe the Const use case is more important. @rpominov's comments above lend credence to this view.

It certainly seems odd to have to go from a type representative to a member of the type (via `of`) in order to get to the empty member (via `empty`). Furthermore, this approach doesn't work for types which satisfy Monoid but not Applicative.


[1]: https://github.com/fantasyland/fantasy-land/pull/162#issuecomment-246651027

safareli:
I tried to make Function a Monoid as in Haskell:
```hs
instance Monoid b => Monoid (a -> b) where
        mempty _ = mempty
        mappend f g x = f x `mappend` g x
```

but we can't:
```js
function.prototype.empty = (x) => empty // what `empty` we what should we return?
function.prototype.concat = function(g) {
  return (x) => this(x).concat(g(x))
}
```

And this problem would be with IO/Task as we don't have a value in it.

--- 

not sure if it would be correct but, as empty is only usable in concat what if had a special partial empty value and check against it in concat. so we can make Identity, Function a monoid again:
```
const isEmpty = e => e['@functional/empty'] === true
// this value is a empty value for any Monoid
const empty = ({
  '@functional/empty': true,
  'constructor': { 'empty': empty},
  'concat': a => a,
})

Function.empty = _ => empty
Function.prototype.concat = function(g) {
  return (x) => {
    const b = g(x)
    if (isEmpty(b)) return this(x)
    else this(x).concat(b)
  }
}

Identity.empty = empty
Identity.prototype.concat = function(b) {
  const a = this.value
  if (typeof a.concat !== 'function') {
    throw new TypeError('`concat` called on Identity which does not hold Monoid value')
  }
  if (isEmpty(b)) return Identity(a)
  else Identity(a.concat(b))
}
```
this way even Id could conform to Monoid [inspiration](http://eighty-twenty.org/2015/01/25/monads-in-dynamically-typed-languages)
(we could have this `isEmpty/empty` in libs like ramda)

p.s.  implementations might not be 100% correct i have not checked them but it gives a general idea. this technique could be used for universal `return` or `of` method which just takes a value and no container. but returns a "partial" value like this:
```js
const of = (a) => ({
  '@functional/of': true,
  value: a,
  chain: (f) => f(a),
  constructor: {'of': of},
  ap: (f) => {
  if (typeof f.constructor.of !== 'function') {
    throw new TypeError('`ap` called on value which is not Applicative')
  }
  return f.constructor.of(a).ap(f)
})
isPartialOf = a => a['@functional/of']
Identity.prototype.ap = (f) => {
  if(isPartialOf(f) return this.map(f.value)
  else Identity(f.value(this.value))
}
```

safareli:
@rpominov @davidchambers  thoughts on this?

rpominov:
@safareli I don't know. It could work but it seems so different from what we've been doing before. Also seems quite complicated.

davidchambers:
I haven't yet wrapped my head around your suggestion, @safareli. It deserves its own thread. :)

I'd love to know what @puffnfresh, @SimonRichardson, @joneshf, or one of the other old hands thinks of this pull request. It's a significant change, but so far it hasn't encountered significant opposition. Is losing the ability to define `fantasy-land/empty` for the Identity type really the biggest downside? Are we overlooking something important?

safareli:
With trick I provided we don't even loos ability to define monoid instance for Identity. Will snippets in my comment and open another thread for that soon!

safareli:
With trick, I provided, we don't even loos ability to define monoid instance for Identity.

Here I have created a [gist](https://gist.github.com/safareli/c7f29f1d7ef02533e47cbb6b6a4e423c) on that idea.

safareli:
[Here](https://github.com/safareli/quasi/tree/initial/test/common) you can check how Function, Identity, Pair, Task could be Monoids. I would need to test them too but it gives an idea of what implementation looks like (it's a bit boilerplate thou). In [this PR](https://github.com/safareli/quasi/pull/1) you can comment/suggest anything ❤️.

safareli:
@puffnfresh, @SimonRichardson, @joneshf any thoughts on this? I think this is exciting change and [safareli/quasi/examples](https://github.com/safareli/quasi/tree/initial/examples) proves that with this changes we lose nothing 🌮

SimonRichardson:
I like it, but want to think about it if you don't mind... Ping me if I
forget to respond..

rpominov:
Wrote some question regarding `quasi` here https://github.com/safareli/quasi/pull/1#issuecomment-252970512 . Probably would be best to move discussion there, this is kinda off-topic for this PR.

davidchambers:
((:bell:)) @SimonRichardson, have you given this matter some thought?

SimonRichardson:
Nope 😱

id_test.js
```diff
@@ -14,17 +14,17 @@ const semigroup = require('./laws/semigroup');
...
 const {of, empty, concat, equals, map} = require('.');
 
 const Id = require('./id');
 
 // Special type of sum for the type of string.
-const Sum = tagged('v');
-Sum[of] = (x) => Sum(x);
-Sum[empty] = () => Sum('');
-Sum.prototype[of] = Sum[of];
-Sum.prototype[empty] = Sum[empty];
```
SimonRichardson: I don't get this change?

davidchambers: At some point I thought `Sum('').constructor` did not evaluate to `Sum`. I was mistaken. I've reverted to the daggy version.

SimonRichardson:
Right, I'm happy with this change, we should bump to v2.0.0 I guess?

rpominov:
+1 from me. I would rather didn't have `empty() -> empty` change though, just because it's not very important but would require major version bump in `static-land`, for instance, because I want it to be as close to `fantasy-land` as possible.

davidchambers:
> I'm happy with this change

Fantastic!

> we should bump to v2.0.0 I guess?

Yes, we should.

> I would rather didn't have `empty() -> empty` change though, just because it's not very important but would require major version bump in `static-land`

Could you elaborate, @rpominov? Is Static Land not affected by the other breaking changes in this pull request?

rpominov:
> Could you elaborate, @rpominov? Is Static Land not affected by the other breaking changes in this pull request?

Yeah, I think so. Not affected by other changes.

rpominov:
One more minor point in favor of method instead of value. In the spec we often need to say something about methods. For example we have a title "Prefixed method names". Strictly speaking we would now have to say something like "methods and values" in all such places which will make it more confusing. Or we could keep saying only "methods" but that also will be confusing because one will think "does this include `empty`?".

davidchambers:
> One more minor point in favor of method over a value. In the spec we often need to say something about methods. For example we have a title "Prefixed method names". Strictly speaking we would now have to say something like "methods and values" in all such places which will make it more confusing. Or we could keep saying only "methods" but that also will be confusing because one will think "does this include `empty`?".

Good point, @rpominov! I believe we should replace the word "method" regardless, as it means different things to different people. In JavaScript I distinguish between "methods" and "functions": the former expect `this` to refer to a value of a particular type; the latter do not. According to this definition, `fantasy-land/empty` is not a method even if we stick with `Monoid m => () -> m`.

We could continue to use "method" in the Java sense of "instance" methods and "static" methods. Regardless of whether we stick with "method" or choose another term, we should define the term.

davidchambers:
@rpominov, I've pushed a commit to rephrase potentially confusing references to "methods". How does it look?

Although I prefer `Monoid m => m` to `Monoid m => () -> m`, specifying a single location for `empty`, `of`, and `chain` is far more important to me. If after weighing the costs of updating Static Land and Static Land libraries against the benefit of a simpler type you feel strongly that we should stick with a nullary function, I'm happy to make that change. Let me know. :)

rpominov:
> @rpominov, I've pushed a commit to rephrase potentially confusing references to "methods". How does it look?

"properties" sounds great 👍 

> If after weighing the costs of updating Static Land and Static Land libraries against the benefit of a simpler type you feel strongly that we should stick with a nullary function, I'm happy to make that change. Let me know. :)

Can't say that I feel strongly, but still feel like we'd better don't make this change. This is not only about static-land and static-land compatible libs. Any fantasy-land@1.0.x compatible lib that didn't delegated Monoid functionality to wrapped items, also won't be affected by the rest of the changes, right?

davidchambers:
I've updated the pull request: `empty` is once again a nullary function. It's now acceptable to refer to "methods", so none of the headings needs to change, so the `<div id="empty-method"></div>` cruft is no longer necessary. :)

If you're happy with the changes, @rpominov, we'll ask @SimonRichardson to sign off and merge. :)

rpominov:
Thank you! Looks great 👍

davidchambers:
Okay. Do you agree, @SimonRichardson? If so, let's merge this and release v2.0.0.

davidchambers:
sanctuary-js/sanctuary-type-classes#6 shows how much simpler this makes interacting with `empty`, `of`, and `chainRec` in a compliant manner. :)

Issue #85, opened 2015-02-06
Why "No parts of b should be checked" for Applicative?


TheLudd:
Why is it important for the Applicative ```of``` function to not check it's input? 

My question arises because I have seen two different types of implementations for ```Maybe```, one that pretty much looks like this where the check to see if the value is null is done in the map function:
```
function Maybe(val) {
  this.val = val;
};
Maybe.prototype.map = function(f) {
  return this.val == null ? this : new Maybe(f(this.val));
};
```
and another one which has 2 sub types ```Just``` and ```Nothing```.
```
function Maybe(val) {
  return this.val == null ? new Nothing() : new Just(val);
};

function Just(val) {
  this.val = val;
};

Just.prototype.map = function(f) {
  return new Maybe(f(this.val));
};

function Nothing() {};

Nothing.prototype.map = function(f) {
  return this;
};
```

My point here is that the second implementation does a check in the constructor, and because of this it would break the fantasy land specification if it implemented ```of = function(val) return Maybe(val)``` but this is not true for the first implementation since it does the checking in the map method.

In the second case, ```Maybe.of``` can never return something that will *not* apply the input function in map.

This leads to the situation that because of internal design decisions ```Maybe.of``` may be forced to be inconsistent with other implementations in order to fulfill this specification. I find this strange and wonder if I have misunderstood something.

SimonRichardson:
That's because you're `Maybe.of` is wrong. It should only return `Just` and not a possibility of both. Checking for null isn't covered in the specification, because this is in a `fantasy-land` and not to mention the [million dollar mistake](http://en.wikipedia.org/wiki/Tony_Hoare#Apologies_and_retractions).

All values should not be null by there very nature when interacting with fantasy-land specification. If you do want to have a null checking part of your `Maybe` may I suggest `Maybe.from` (or similar named). There for your `of` can follow the correct specification.

```
Maybe.from = function(x) {
    return x == null ? new Nothing() : new Just(x);
}
```

TheLudd:
@SimonRichardson Is the first implementation of Maybe also wrong?

Perhaps I don't understand the use case for ```of```. If all it does is wrap the value in a ```Just``` why wrap it at all? 
I'd be grateful for an explanation of how/when to use ```of```.

michaelficarra:
Wrapping puts the value in a context so that you know which `map` to apply.

SimonRichardson:
[...]

PR #184, opened 2016-10-16
Enforce parametricity

rpominov:
We already have similar note for `of` method, but I figured we should add it for other methods as well.

See also https://github.com/fantasyland/fantasy-land/issues/85


davidchambers:
:deciduous_tree:

rpominov:
Ok, with approval from @davidchambers , @joneshf , and @rjmk I think it's safe to merge.

PR #192, opened 2016-10-23
Another go at updating dependencies

joneshf:
Re: #160


README.md
```diff
@@ -25,7 +25,7 @@ structures:
 * [Bifunctor](#bifunctor)
 * [Profunctor](#profunctor)
 
-<img src="figures/dependencies.png" width="677" height="212" />
+<img src="figures/dependencies.png" />
```
rjmk: Could we use the svg rather than the png here?

joneshf: Could just kill the png altogether.

figures/dependencies.dot
```diff
@@ -0,0 +1,35 @@
+digraph {
+  node [fontcolor=blue;shape=plaintext]
```
davidchambers: Let's `s/blue/#4078c0/` to match GitHub's link colour.

README.md
```diff
@@ -25,7 +25,7 @@ structures:
 * [Bifunctor](#bifunctor)
 * [Profunctor](#profunctor)
```
davidchambers: Let's remove this list. It's duplicative now that the diagram is up to date. :)

davidchambers:
@joneshf, what command are you running to generate __dependencies.svg__? Could you add it to a makefile so the rest of us know how to do it?

README.md
joneshf: I think it's best to keep this. It's nice to have a fallback in case something happens with the image, different people process information differently, and it's still plaintext so it can be searched.

davidchambers: Two points in favour of the ASCII version. :stuck_out_tongue_winking_eye:

README.md
```diff
@@ -25,7 +25,7 @@ structures:
...
 
-<img src="figures/dependencies.png" width="677" height="212" />
+<img src="https://cdn.rawgit.com/fantasyland/fantasy-land/master/figures/dependencies.svg" />
```
joneshf: Ah! Makes sense.

joneshf:
This file: https://github.com/fantasyland/fantasy-land/blob/4c1e3ee239c9a68e8d0a9a299b663e48d4cb6cfe/figures/dependencies.sh

davidchambers:
The links in the image do not work, because GitHub wraps the image in an anchor element. Shall we remove the links? Since we're going to keep the list of section links they're not essential.

joneshf:
Sure.

README.md
joneshf: :smile: 

I'm fine with the output graph being ASCII. I just don't want to maintain the shape of the graph in raw ASCII.

davidchambers:
It's looking good! Ideally we'd now have two clean commits:

__Commit 1:__

  - add __figures/dependencies.dot__
  - add __figures/dependencies.sh__
  - update __figures/dependencies.svg__

__Commit 2:__

  - update __README.md__ to point to RawGit SVG (include hash of Commit 1)
  - remove __figures/dependencies.mp__
  - remove __figures/dependencies.png__

I'm happy to do this if fiddling with `git rebase` is not the way you like to spend your Sundays. :)

:warning: We'll then merge the pull request **without** using the "squash and merge" feature.

README.md
davidchambers: I'm totally with you. A program like `dot` which generates an ASCII diagram would be neat!

I very much like your change. Thanks for taking the time to work on it.

joneshf:
Nah, rebase is fine.

I wonder if svg is worth it if we have to use rawgit like this. Remembering to make one commit to change the graph and one to update the sha is less than ideal. If we use png rather than svg for output, it seems like we could do it in one commit, and similar for future changes. What do you think?

davidchambers:
Good point. Let's stick with a PNG. :)

README.md
joneshf: It does support multiple output formats, including `plain`. `plain` gives you the size of the graph, a sequence of nodes with positions, and a sequence of edges with position if you want to parse the coordinates into an ascii thing. It looks like a regular language, so probably not too hard to parse.

README.md
davidchambers: @rpominov added the dimensions in #119 for a good reason. We should specify the new dimensions to avoid regressing.

README.md
davidchambers: Cool! I may investigate that one day (but certainly not today).

davidchambers:
Excellent! If you're happy with the current state of the diff, Hardy, let's squash and merge. :)

SimonRichardson:
👍

PR #193, opened 2016-10-25
Fix id_test and argument order in laws

safareli:
related to #173 #171

### DIFF
--- a/id_test.js
+++ b/id_test.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const fl = require('.');
+const fl = require('./');
 const applicative = require('./laws/applicative');
 const apply = require('./laws/apply');
 const chain = require('./laws/chain');
--- a/id_test.js
+++ b/id_test.js
@@ -35,7 +35,7 @@ Sum.prototype[equals] = function(x) {
 
 const equality = (x, y) => x[equals] ? x[equals](y) : x === y;
 const test = f => t => {
-  t.ok(f('x'));
+  t.ok(f('x') === true);
   t.done();
 };
 
--- a/id_test.js
+++ b/id_test.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const fl = require('.');
 const applicative = require('./laws/applicative');
 const apply = require('./laws/apply');
 const chain = require('./laws/chain');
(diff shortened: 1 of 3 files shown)


id_test.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const fl = require('.');
```
davidchambers: Could you change the path to `'./'`? `'.'` has always worked for me, but I've seen it fail for other people.

davidchambers:
It's pretty clear no one is using these at the moment. ;)

safareli:
i'm using it  but just updated from v1 to v2

rpominov:
I wonder if maybe these should live in a separate project. Here are couple benefits:

 - No extra deps #168
 - The version number of `fantasy-land` package will only track changes in the spec itself.

davidchambers:
> The version number of `fantasy-land` package will only track changes in the spec itself.

I find this compelling. It'd be a shame to keep incrementing the spec's version in order to release fixes such as these, but it'd be unreasonable to force such fixes to wait until a new version of the spec is released. Allowing the two projects to be released independently strikes me as a good idea.

safareli:
I think changes like this should bump patch version for now.

davidchambers:
:deciduous_tree:

SimonRichardson:
I still think the laws should be in this repo, but I think they should be dramatically improved (or reworked, refactored, re-written). They're very old, but are worth their gold when they did work. Personally we should go one step further and make them links in the spec, so people are more likely to know about them.

rpominov:
I think laws should look like this:

```js
// `u.map(a => a)` is equivalent to `u` (identity)
const identity = (u) => {
  return {
    left: u[map](x => x),
    right: u,
  }
}

// `u.map(x => f(g(x)))` is equivalent to `u.map(g).map(f)` (composition)
const composition = (u, f, g) => {
  return {
    left: u.map(x => f(g(x))),
    right: u.map(g).map(f),
  }
}
```

When we're trying to make them smarter, they just end up not complete or hard to use with certain types.

Also this simple laws implementations will be easy to keep up to date because it's basically copy-paste of the snippets from spec.

I can make a PR that simplifies laws like this if this sounds like a good idea.

safareli:
```
const identity = (u) => {
  return {
    left: u[map](x => x),
    right: u,
  }
}
{left,right} = identity(Maybe.Just(1))
maybeEquals(left,right)
```

i don't see a big difference with this
```
const identity = (eq, u) => eq( u[map](x => x), u)
identity(maybeEquals, Maybe.Just(1))
```
and imo the last one is better.

rpominov:
This refactoring will be a breaking change strictly speaking though.

safareli:
yeh :/ that's true.

What if we state in readme that we might have some braking changes in laws, but still do patch releases if there is no change in actual spec.

rjmk:
Automated tools don't read READMEs. I think we need to stick to incrementing the major version for a breaking change

rpominov:
Let's move this discussion to #195

safareli:
despite the #195 can we merge this PR?

PR #196, opened 2016-10-27
Use uppercase letters for TypeRep

safareli:
In Applicative we are using `A` but in monad and monoid we still use `m`.
in case of Monad one might think that we are accessing `of` on instance.
monoid law is technically correct but I think sticking to one way of doing thing is better so i have updated it as well.


safareli:
squished and updated

davidchambers:
:deciduous_tree:

PR #197, opened 2016-10-27
Add Alt, Plus and Alternative specs

### LINKED ISSUE #187: Add spec for Alt, Plus, Alternative, MonadZero, MonadOr and MonadPlus
I think it would be nice if we also define algebras for `Alt`, `Plus` and `Alternative`

``` hs
-- | The `Alt` type class identifies an associative operation on a type
-- | constructor.  It is similar to `Semigroup`, except that it applies to
-- | types of kind `* -> *`, like `Array` or `List`, rather than concrete types
-- | `String` or `Number`.
-- |
-- | `Alt` instances are required to satisfy the following laws:
-- |
-- | - Associativity: `(x <|> y) <|> z == x <|> (y <|> z)`
-- | - Distributivity: `f <$> (x <|> y) == (f <$> x) <|> (f <$> y)`
-- |
-- | For example, the `Array` (`[]`) type is an instance of `Alt`, where
-- | `(<|>)` is defined to be concatenation.
class Functor f <= Alt f where
  alt :: forall a. f a -> f a -> f a


-- | The `Plus` type class extends the `Alt` type class with a value that
-- | should be the left and right identity for `(<|>)`.
-- |
-- | It is similar to `Monoid`, except that it applies to types of
-- | kind `* -> *`, like `Array` or `List`, rather than concrete types like
-- | `String` or `Number`.
-- |
-- | `Plus` instances should satisfy the following laws:
-- |
-- | - Left identity: `empty <|> x == x`
-- | - Right identity: `x <|> empty == x`
-- | - Annihilation: `f <$> empty == empty`
class Alt f <= Plus f where
  empty :: forall a. f a


-- | The `Alternative` type class has no members of its own; it just specifies
-- | that the type constructor has both `Applicative` and `Plus` instances.
-- |
-- | Types which have `Alternative` instances should also satisfy the following
-- | laws:
-- |
-- | - Distributivity: `(f <|> g) <*> x == (f <*> x) <|> (g <*> x)`
-- | - Annihilation: `empty <*> f = empty`
class (Applicative f, Plus f) <= Alternative f
```

Snippets from [purescript-control](https://github.com/purescript/purescript-control/tree/master/src/Control)

It will be useful to `Future/Task` implementations (`alt` will be `race`).
and with also define `Parallel` algebra then in concurrent applicative could implement `Alternative` and we would have the `race` as well (`MonadRace` is removed).

safareli:
Specs added:
- Alt
- Plus
- Alternative
- <s>MonadZero</s>
- <s>MonadOr</s>
- <s>MonadPlus</s>

TODO:
- [x] add laws in `laws/*`
- [x] update test
- <s>add `mPlus` and `mOrElse` ?</s>

fixes #187

Monad{Zero,Or,Plus} are removed from this PR ([reason](https://github.com/fantasyland/fantasy-land/pull/197#issuecomment-257571690))


README.md
```diff
@@ -243,6 +249,69 @@ Given a value `f`, one can access its type representative via the
...
+
+1. `pempty` must return a value of the same Plus
+
+### Alternative
+
+A value that implements the Alternative specification must also implement
+the [Applicative](#applicative) and [Plus](#plus) specifications.
+
+1. `x.ap(f.alt(g))` is equivalent to `x.ap(f).alt(x.ap(g))` (distributivity)
+1. `M.pempty().ap(f)` is equivalent to `M.pempty()` (annihilation)
```
safareli: In purescript it's in different order and i think it's correct, so I have opened issue [purescript-control#34](https://github.com/purescript/purescript-control/issues/34)

safareli:
Also we can use `zero` instead of `pempty` (from [semigroupoids.Plus](https://hackage.haskell.org/package/semigroupoids-5.0.0.1/docs/Data-Functor-Plus.html))

garyb:
👍 for `zero` - we don't use that to avoid the `mempty` situation in PureScript only because it's used for yet another "zero" - that of `Semiring` :smile:

safareli:
Will change them! and if in future we add `Semiring` to FL, as it is subclass of `Monoid` than we would just use monoids `empty` for it.

garyb:
Well, this is a discussion for another day, but it's not _really_ a subclass of `Monoid` as there are two possible `Monoid`s for it: `(+)` and 0, and `(*)` and 1. We have `Additive` and `Multiplicative` newtypes for those `Monoid`s for any `Semiring`... not sure what the JS answer to that would be!

rpominov:
The PR LGTM overall 👍

internal/maybe.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+Maybe[fl.zero] = () => Maybe.Nothing;
+
+Maybe.prototype[fl.alt] = function(b) {
+  return this.cata({
+    Just: _ => this,
+    Nothing: () => b,
+  });
+};
```
SimonRichardson: I wonder if we just import `fantasy-options`, I know people don't want more dependencies, but I don't want to envision lots of `Maybe`s out in the wild. Everyone different to each other...

laws/alt.js
```diff
@@ -0,0 +1,24 @@
...
+  a[alt](b)[alt](c),
+  a[alt](b[alt](c))
+);
+
+const distributivity = eq => a => b => f => eq(
+  a[alt](b)[map](f),
+  a[map](f)[alt](b[map](f))
+);
+
+module.exports = {associativity, distributivity};
```
SimonRichardson: 👍

laws/alternative.js
```diff
@@ -0,0 +1,24 @@
...
+  x[ap](f[alt](g)),
+  x[ap](f)[alt](x[ap](g))
+);
+
+const annihilation = T => eq => x => eq(
+  x[ap](T[zero]()),
+  T[zero]()
+);
+
+module.exports = {distributivity, annihilation};
```
SimonRichardson: 👍

laws/monador.js
```diff
@@ -0,0 +1,18 @@
...
+1. `M.of(a).alt(b)` is equivalent to `M.of(a)` (left catch)
+
+**/
+
+const leftCatch = T => eq => a => b => eq(
+  T[of](a)[alt](b),
+  T[of](a)
+);
+
+module.exports = {leftCatch};
```
SimonRichardson: 👍

laws/monadplus.js
```diff
@@ -0,0 +1,18 @@
...
+1. `x.alt(y).chain(f)` is equivalent to `x.chain(f).alt(y.chain(f))` (distributivity)
+
+**/
+
+const distributivity = eq => x => y => f => eq(
+  x[alt](y)[chain](f),
+  x[chain](f)[alt](y[chain](f))
+);
+
+module.exports = {distributivity};
```
SimonRichardson: 👍

laws/monadzero.js
```diff
@@ -0,0 +1,18 @@
...
+1. `M.zero().chain(f)` is equivalent to `M.zero()` (annihilation)
+
+**/
+
+const annihilation = T => eq => f => eq(
+  T[zero]()[chain](f),
+  T[zero]()
+);
+
+module.exports = {annihilation};
```
SimonRichardson: 👍

laws/plus.js
```diff
@@ -0,0 +1,30 @@
...
+  x[alt](T[zero]()),
+  x
+);
+
+const annihilation = T => eq => f => eq(
+  T[zero]()[map](f),
+  T[zero]()
+);
+
+module.exports = {rightIdentity, leftIdentity, annihilation};
```
SimonRichardson: 👍

SimonRichardson:
Just one [comment](https://github.com/fantasyland/fantasy-land/pull/197#pullrequestreview-6317474) otherwise 🎉

internal/maybe.js
safareli: That's interesting point, suppose we depend on fantasy-id, fantasy-options, fantasy-compose. what would happen if someone adds other spec?

safareli:
That's interesting point, suppose we depend on fantasy-id, fantasy-options, fantasy-compose ... 

What would happen if someone adds other spec? how do you see for example process of adding some spec `Foo` with method `foo`?

Also `fantasy-*` libs have test on there own, we would test it twice? and they also depend on FL.

safareli:
Possibly solution is to completely remove `test` from FL, but in this case fantasy-id, fantasy-maybe ... should be always up to date and if we have a PR to add some spec to FL before we land it PRs for `fantasy-*` should be opened to ensure PR in FL is correct, then PR in FL could be merged and version bumped. After which PRs in `fantasy-*` could update FL and merge as well.

internal/maybe.js
safareli: Here is [my responce](https://github.com/fantasyland/fantasy-land/pull/197#issuecomment-257029476) to this issue (scroll down)

joneshf: I'm sorry. I was venting. This PR is not the appropriate place for it. Do not let my comment stop progress of this PR. If it's something we are going to address, we should do it in another issue.

SimonRichardson: I feel the same way though :-(

safareli:
one issue is that i Maybe is not MonadPlus but is MonadOr
```hs
mplus a b >>= f = mplus (a >>= f) (b >>= f)
a = Just 0
b = Just 1
f 0 = Nothing
f 1 = Just 1
```
so this will fail:
`js
test(x => monadPlus.distributivity(equality)(Maybe.Just(0))(Maybe.Just(1))(a => a === 0 ? Maybe.Nothing : Maybe.Just(a)))
```

but Array is MonadPlus not MonadOr so will update pr

**Comment by safareli** [comment]
Not ready to merge, there is one issue, I would post about it tomorrow

**Comment by safareli** [comment]
At this point it's ready to be merged, **But** there is one thing, which I don't quite like:

If some structure has `of` we know that it's Apply and if it also has `chain` than it must be a Monad there are no other choices. on the other hand if structure has `of` , `chain` and `alt` than it must be `MonadZero`. Problem is that it could also be a `MonadPlus` or `MonadOr`. so we have 3 specs `Monad{Zero,Plus,Or}` and they are not distinguishable from each other. 

I think it could be problematic in js (because of lack of type system), but it could also be problematic even in purescript?
```hs
f :: (Monad___ m) => m a ->  m a -> (a-> m b) -> m b
f m1 m2 g= (m1 `alt` b2) >>= g
```

I don't have strong opinion on it, maybe it's not a big dial and it's actually ok, but wanted to point it out. Also I would love to hear what @garyb and @paf31 think on it as in PS there is not MonadOr (yet).

joneshf:
> If some structure has of we know that it's Apply and if it also has chain than it must be a Monad there are no other choices. on the other hand if structure has of , chain and alt than it must be MonadZero

Is that true? It seems like you can have something tha is `Applicative` and `Chain`, but doesn't satisfy the `Monad` laws. Similarly for `MonadZero`.

safareli:
@joneshf technically one could make `Validation` a `Chain` but it's sort of cheating.

So to be precise if structure is Applicative(of,ap) and Chain(chain) than we have two choices
- Applicative+Chain
- Applicative+Chain+Monad

if structure is Applicative(of,ap) Alternative(alt,zero) and Chain(chain) then we have 4 choices:
- Applicative+Alternative+Chain
- Applicative+Alternative+Chain+MonadZero
- Applicative+Alternative+Chain+MonadZero+MonadOr
- Applicative+Alternative+Chain+MonadZero+MonadPlus

rpominov:
I wouldn't care too much about detection of supported algebras by presence of methods. If this will become a real problem we can solve it for example by requiring values to have `fantasy-land/algebras` that point to an array like `['Funcor', 'Apply', 'Applicative']`.

But we should think about structures that may support both MonadPlus and MonadOr.

safareli:
> But we should think about structures that may support both MonadPlus and MonadOr, as far as I understand.

Exactly.

In that case we should have separate methods for MonadPlus and MonadOr. in the haskell proposal names: `mplus` and `morelse` are used, if we use them I think they should be camelCase: `mPlus`, `mOrElse`.

if structure is MonadPlus or MonadOr it could define `alt` using `mPlus` or `mOrElse`.
```js
T.prototype[fl.mOrElse] = T.prototype[fl.alt] = function(a) { .... }
T.prototype[fl.mPlus] = T.prototype[fl.alt] = function(a) { .... }
```

do others agree on the names?

rpominov:
Maybe just `plus` and `orElse`?

rpominov:
Is there an example of a type that supports both MonadPlus and MonadOr but with different `mPlus` and `mOrElse` methods? Just trying to understand.

Also just a though, maybe completely invalid: maybe it's too early to add MonadPlus and MonadOr to Fantasy Land? As I understand this stuff is pretty new, and maybe we should wait. We could open an issue for it and see if people come up with use cases for these specs. Just one of concerns: we reserve method names for this in our namespace, which we might want to use for other stuff in the future.

safareli:
You can see example above for Maybe and []

safareli:
About usefulness Alternative normalised monoidal nature of Task/Future

for `Task ()`
- Task.zero() = Task.rejected()
- Task.alt = Task.mOrElse = (a) => [`this.orElse(() => a)`](https://github.com/rpominov/fun-task/blob/master/docs/api-reference.md)

for TaskAp
- Task.zero() = Task.empty() a.k.a. Task.never()
- Task.alt = (a) => [`Task.race([this,a])`](https://github.com/rpominov/fun-task/blob/master/docs/api-reference.md#taskorelsefn)

rpominov:
> As described above, Array and Maybe support both:

Ah, sorry, missed that.

> `Task.alt = Task.mOrElse (is MonadOr) = (a) => this.orElse(() => a)`

Pity that if we use `orElse` as method name instead of `mOrElse`, there will be a problem for some Task/Future implementations :)

Regarding usefulness, I don't have a big enough perspective for all this stuff to judge. To me it just seems than maybe we move too fast for a specification on which a lot of code may depend, so future changes may be hard. But as I said, I'm a bad person to judge.

safareli:
yah `orElse` is used in folktale's Either, Validation, Maybe and Task (@robotlolita) and in fun-task too, Fluture (@Avaq) has `or` but it's more Applicative Alt. So as `orElse` is used so widely maybe have `m` prefix?

This change is not braking, there are many such Monoidal structures, which could be Alternative and by adding the spec users of those structures will have much better experience I think. 

Also by being able to express `race/or` using this algebras will make possible to implement some interesting Free structures, imagine Free MonadZero and Free Alternative, I believe they would be pretty usefull.

rpominov:
Yea, it's just we so far successfully avoided to use prefixes like `fmap` vs `map` etc. If we use prefix should it be full word `monad` similarly to `chain` in `chainRec`?

joneshf:
> To me it just seems than maybe we move too fast for a specification on which a lot of code may depend, so future changes may be hard. But as I said, I'm a bad person to judge.

This is a welcome change to the previous three years.

joneshf:
> So to be precise if structure is Applicative(of,ap) and Chain(chain) than we have two choices
> 
> Applicative+Chain
> Applicative+Chain+Monad
> if structure is Applicative(of,ap) Alternative(alt,zero) and Chain(chain) then we have 4 choices:
> 
> Applicative+Alternative+Chain
> Applicative+Alternative+Chain+MonadZero
> Applicative+Alternative+Chain+MonadZero+MonadOr
> Applicative+Alternative+Chain+MonadZero+MonadPlus

I don't think we can choose that. The data type tells you whether it's possible or not. I don't have an example off hand of something that is `Applicative` and `Chain`, but not `Monad`. However, I find it hard to think that said data type cannot exist. So I don't know that we can always assume that if something has `of` and `chain` that it's automatically a `Monad`.

safareli:
@joneshf yes that's what i ment 
> ...than we have two choices
- Applicative+Chain (structure is just Applicative and Chain)
- Applicative+Chain+Monad (or it's also a Monad)

but as @rpominov also pointed out it's not only about 4 choices it's about being able to express both in one structure, for example Maybe and [] could be both MonadOr and MonadPlus. and also users of the structures being sure how the method will behave.


btw @davidchambers I think in `sanctuary-type-classes` if structure has `of/chain` we assume that it's a Monad.

safareli:
About names I don't know what's our strategy here. The method name `chainRec` [was proposed](https://github.com/fantasyland/fantasy-land/issues/151#issuecomment-244249844) by @scott-christopher, but now I think maybe it would have been better if we used `tailRec` as now I don't think of any *Rec except ChainRec. but the full prefix in this case `.monadOrElse`, `.monadPlus` seems to clunky. they do not have collision between methods in FL it self. What's the reason of namespaced methods if we don't use method name because it's used in some other libs?

rpominov:
> They do not have collision between methods in FL it self

They may have collision with names of methods that we may want to add in the future though. So we should be careful to not pick too general names. What if we name methods after type classes:

```
Chain -> chain
ChainRec -> chainRec
MonadZero -> monadZero
MonadPlus -> monadPlus
MonadOr -> monadOr
```

?

safareli:
I thought about `Monad{Or,Plus,Zero}` stuff and it's really ambiguous, like we have multiple variations of laws, it's not that standardized:
- https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#t:MonadPlus
- https://wiki.haskell.org/MonadPlus
- https://wiki.haskell.org/MonadPlus_reform_proposal
- https://wiki.haskell.org/All_About_Monads#Zero_and_Plus
- https://en.wikibooks.org/wiki/Haskell/MonadPlus#The_MonadPlus_laws
- https://mail.haskell.org/pipermail/libraries/2014-February/022004.html
- https://www.reddit.com/r/haskell/comments/1x5bvf/what_the_heck_is_a_rightseminearring_also/
- http://comonad.com/reader/wp-content/uploads/2009/07/AllAboutMonoids.pdf (the last slide)
    
So i think best way to go is to remove `Monad{*}` stuff from this PR and only introduce Alternative, Plus and Alt. This will be immediately useful in for example Applicative Task/Futures, which do not exist **yet**, after there are such Alternative structures we can make cool Free structures ^_^

safareli:
### I think it's ready for final review and then it could be merged 🎉  🌮

README.md
```diff
@@ -25,7 +28,7 @@ structures:
 * [Bifunctor](#bifunctor)
 * [Profunctor](#profunctor)
 
-<img src="figures/dependencies.png" width="863" height="347" />
+<img src="figures/dependencies.png" width="100%" />
```
davidchambers: I think this will reintroduce the problem @rpominov fixed in #119.

README.md
```diff
@@ -243,6 +246,69 @@ Given a value `f`, one can access its type representative via the
...
+the [Alt](#alt) specification.
+
+1. `x.alt(A.zero())` is equivalent to `x` (right identity)
+2. `A.zero().alt(x)` is equivalent to `x` (left identity)
+2. `A.zero().map(f)` is equivalent to `A.zero()` (annihilation)
+
+#### `zero` method
+
+```hs
+zero :: Plus x => () -> x
```
rpominov: Should it be `Plus x => () -> x a`? Also maybe use `p` instead of `x`?

README.md
safareli: will fix that we definitly need CONTRIBUTING.md

README.md
safareli: In that case we should also add `forall a` I think, but i'm not familiar with that. maybe @joneshf could help here?

rpominov: I think `forall` is kinda implicit in our loosely specified type system. What I was trying to say is that `x` is not a type here, it's a type constructor, so it cannot be in that position in signature. `x a` is a type on the other hand.

safareli: I would use `f a`. `f` to be consistent with other types

README.md
safareli: on github the width of README is fixed but for example on [npm](https://www.npmjs.com/package/fantasy-land) it looks bad because max width of readme area is 727px and if you are on small screan than it's worth.

rpominov: Yeah, same issue with mobile version of GitHub for example. I wish we could find a fix without this side effect, but in the current situation having links to point to the right locations is more important imo.

rpominov: +1 for moving to the end.

safareli: you can check rendered image [here](https://github.com/safareli/fantasy-land/tree/alternative)

safareli:
@SimonRichardson @davidchambers I think it's ready to be merged ⏲

SimonRichardson:
I'll check tomorrow if that's ok

safareli:
⏰ a friendly reminder

SimonRichardson:
@safareli sorry, I looked at this yesterday and forgot to 👍 

MERGE IT
🌮

davidchambers:
Would you mind squashing the commits, @safareli? It would be nice to avoid mentioning MonadZero, MonadOr, and MonadPlus in the commit history to avoid confusion.

safareli:
I think this commits might be useful in future for reference. And as I remember before my committee where `Squash and Merged`. can we do it again? if yes then commit message will be 

```
Add Alt, Plus and Alternative specs (#197)
```

And for it's description we could just say that tests where refactored a bit:
```
- refactor tests
- 'id.js' moved to 'internal/'
```

if you want to just `Rebase and Merge` I could squash changes but the monad zero... stuff will be lost.

davidchambers:
Thanks for updating the pull request's title and providing an appropriate commit message. :)

Shall we release v2.1.0?

safareli:
I think yes 🎈

rpominov:
Yea, releasing a new version would be nice. I need it to add this algebras to static-land.

Issue #199, opened 2016-10-30
Move all functions to namespaced Type Representative


briancavalier:
Recently, I started updating two libs ([creed](https://github.com/briancavalier/creed) and [most](https://github.com/cujojs/most)) to support FL 2.0, and added Static Land support to most.  In the process, I read [this thread](https://github.com/fantasyland/fantasy-land/pull/180) that led to standardizing on Type Representatives for static functions like `of` and `empty`.

All of that led me to the thought of leveraging Type Representatives more fully.  I wanted to put the idea out for discussion to see if it seems interesting:

Rather than many namespaced methods, e.g. `x[fl.map](f)`, FL could move to functions on a single namespaced Type Representative, e.g. `x[fl].map(f, x)`, where `x[fl]` is the location of the `x`'s Type Representative, and `fl` is the only required FL string/Symbol.

This is obviously a substantial change, but I think it would have several benefits:

1. It standardizes on functions, rather than a mix of functions and methods.
2. It reduces the number of namespaced properties on instances to exactly 1, allowing functions on the Type Representative to be non-namespaced (though they *could* remain namespaced if that's desirable)
3. It removes the need to use `.constructor` as the location of the Type Representative.  Although, it may also be worth considering whether it's useful to add the Type Representative as `Constructor[fl]` for cases where where no instance is available.
4. While it doesn't completely unify FL and Static Land, it does come close:
  - this new, richer Type Representative is effectively a Static Land dictionary, and
  - provides a standard location for Static Land dictionaries.  When an instance is available, it reduces the need to explicitly pass Static Land dictionaries around.

I don’t think this has many downsides compared to the current FL 2.0 approach.  For example, ergonomics: the prefixed methods required by FL 2.0 are, imho, already fairly unergonomic for app devs.  Library devs will endure them in order to get the benefits of being FL compliant.  Moving to functions on a single namespaced type representative seems like it retains the same level of ergonomics.

Any FL delegation library such as fantasy-sorcery or Ramda would need to change its dispatching, which would be a breaking change, but the dispatching is no more complex.

What do you think of the idea of a all functions on a single namespaced Type Representative?

rpominov:
Sounds like a great idea! If we do this right this will effectively make fantasy-land as expressive as static-land while kipping all use-cases of fantasy-land. Here is how we could do this: 

[...]

PR #206, opened 2016-11-29
traverse: Apply -> Applicative

tycho01:
afaict this was a typo. very confusing for learners like me. :<

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ Compose.prototype.map = function(f) {
 #### `traverse` method
 
 ```hs
-traverse :: Apply f, Traversable t => t a ~> (a -> f b, c -> f c) -> f (t b)
+traverse :: Applicative f, Traversable t => t a ~> (a -> f b, c -> f c) -> f (t b)
 ```
 
 A value which has a Traversable must provide a `traverse` method. The `traverse`


safareli:
Yes it's a typo
>     2. `f` must return a value of an Applicative

rpominov:
See also https://github.com/fantasyland/fantasy-land/pull/147/files#r79309177

safareli:
@rpominov as you pointed out maybe we should change laws so it's consistent with type signature?

Or we could change the method to something like this?
```
traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
```

Or we need to add FAQ.md 😄

rpominov:
Yea, +1 to at lest make it consistent.

safareli:
For `Tuple z a` user could provide function of form `a ->  Tuple z a`  (for example `(a) => Tuple([], a)`) and I think it would work properly. So we should change `Applicative` to `Apply` in law?
or use `TypeRep f` instead of passing `a -> f a`? @SimonRichardson @davidchambers @scott-christopher

scott-christopher:
As per the comment @rpominov linked to above, my original thoughts were to use the weaker constraints of `Apply` given that `of` had to be explicitly passed anyway meaning it was already described in the signature.

Now that we have the language to describe `Applicative f => TypeRep f`, it could make it clearer to ask for the full `TypeRep` now rather than just the `of` function.

rpominov:
> Now that we have the language to describe Applicative f => TypeRep f, it could make it clearer to ask for the full TypeRep now rather than just the of function.

That a good idea for the future. Btw this is how it's done in static-land, so we should get this automatically if [unification](https://github.com/fantasyland/fantasy-land/issues/199) happen.

But for now we should at least fix the inconsistency: either make a change in the signature or in laws.

scott-christopher:
> But for now we should at least fix the inconsistency: either make a change in the signature or in laws.

Sure, in that case I'm personally happy with this PR as it is.

davidchambers:
> Now that we have the language to describe `Applicative f => TypeRep f`, it could make it clearer to ask for the full `TypeRep` now rather than just the `of` function.

I'm in favour of this change. It's beyond the scope of this pull request, though.

safareli:
@fantasyland/core if others think this is the way to go please approve this.

joneshf:
Thanks so much @tycho01 and every involved!

PR #213, opened 2016-12-16
update Extend and Comonad dependencies to match figure

davidchambers:
Thank you for pointing out this inconsistency, @jlmorgan.

If Extend really *does* depend on Functor we should update the description of Extend rather than merge this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ structures:
 * [Bifunctor](#bifunctor)
 * [Profunctor](#profunctor)
 
-<img src="figures/dependencies.png" width="888" height="340" />
+<img src="figures/dependencies.png" width="888" height="347" />
 
 ## General
 
@@ -457,6 +457,8 @@ the [Applicative](#applicative) and [Chain](#chain) specifications.
 
 ### Extend
 
+A value that implements the Extend specification must also implement the [Functor](#functor) specification.
+
 1. `w.extend(g).extend(f)` is equivalent to `w.extend(_w => f(_w.extend(g)))`
 
 #### `extend` method
@@ -481,11 +483,10 @@ method takes one argument:
 
 ### Comonad
 
-A value that implements the Comonad specification must also implement the [Functor](#functor) and [Extend](#extend) specifications.
+A value that implements the Comonad specification must also implement the [Extend](#extend) specification.
 
 1. `w.extend(_w => _w.extract())` is equivalent to `w`
 2. `w.extend(f).extract()` is equivalent to `f(w)`
-3. `w.extend(f)` is equivalent to `w.extend(x => x).map(f)`
 
 #### `extract` method
 
--- a/laws/comonad.js
+++ b/laws/comonad.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const {identity} = require('fantasy-combinators');
-const {extend, map, extract} = require('..');
+const {extend, extract} = require('..');
 
 /**
 
@@ -9,7 +9,6 @@ const {extend, map, extract} = require('..');
 
 1. `w.extend(_w => _w.extract())` is equivalent to `w`
 2. `w.extend(f).extract()` is equivalent to `f(w)`
-3. `w.extend(f)` is equivalent to `w.extend(x => x).map(f)`
 
 **/
 
@@ -25,10 +24,4 @@ const rightIdentity = t => eq => x => {
   return eq(a, b);
 };
 
-const associativity = t => eq => x => {
-  const a = t(x)[extend](identity);
-  const b = t(x)[extend](identity)[map](identity);
-  return eq(a, b);
-};
-
-module.exports = {leftIdentity, rightIdentity, associativity};
+module.exports = {leftIdentity, rightIdentity};
--- a/test.js
+++ b/test.js
@@ -67,7 +67,6 @@ exports.chainRec = {
 exports.comonad = {
   leftIdentity: test(comonad.leftIdentity(Id[fl.of])(equality)),
   rightIdentity: test(comonad.rightIdentity(Id[fl.of])(equality)),
-  associativity: test(comonad.associativity(Id[fl.of])(equality)),
 };
 
 exports.extend = {


SimonRichardson:
👍

davidchambers:
Let's wait for confirmation from @joneshf before merging this pull request (he added the relationship in #192).

rpominov:
But shouldn't Comonad depend on Functor? Spec says so.

davidchambers:
Thanks, @rpominov! I've updated the pull request. :)

safareli:
in PS [`class Functor w <= Extend w `](https://github.com/purescript/purescript-control/blob/master/src/Control/Extend.purs#L24)

davidchambers:
That's instructive, @safareli. Shall we revert the changes to the diagram and instead move the reference to the Functor dependency in the spec from Comonad to Extend?

safareli:
>  Shall we revert the changes to the diagram and instead move the reference to the Functor dependency in the spec from Comonad to Extend?

`Extend` is dual of `Chain` which is a Functor, also `Extend` holds values (it's kind is `* -> *`)  so I think it should be a Functor, but lets see what @joneshf thinks.

joneshf:
Yeah, we do want `Functor -> Extend`. We should make that explicit in the definition and move `Comonad`s third law (`w.extend(f)` is equivalent to `w.extend(x => x).map(f)`) to `Extend`.

README.md
```diff
@@ -28,7 +28,7 @@ structures:
 * [Bifunctor](#bifunctor)
 * [Profunctor](#profunctor)
 
-<img src="figures/dependencies.png" width="888" height="340" />
+<img src="figures/dependencies.png" width="888" height="347" />
```
davidchambers: This is now an unrelated—but still intentional—change.

davidchambers:
Please have another look, @joneshf.

safareli:
btw the law with `map` is not present in PS  [Extend](https://github.com/purescript/purescript-control/blob/master/src/Control/Extend.purs#L23) [Comonad](https://github.com/purescript/purescript-control/blob/master/src/Control/Comonad.purs#L18-L19) so do we have some special need for that?

jlmorgan:
Is there a way to implement `Extend` using `map` like how `Chain` can be implemented using:

```javascript
chain(f) { return this.map(f).join(); }
join() { return this.value; }
```

safareli:
@joneshf dont know we should take a look at variations of laws in haskell [Comonad](https://hackage.haskell.org/package/comonad-5/docs/Control-Comonad.html#v:-61--62--61-). looks like the map law we have is close to default definition of `extend`
```hs
extend f = fmap f . duplicate
```
if we substitute `duplicate` with `extend id`, we get:
```js
extend f = fmap f . extend id
```

joneshf:
@jlmorgan Sure.

```js
extend(f) { return this.duplicate().map(f); }
duplicate() { return F(this); }
```

@safareli That law is only needed if you're defining both `duplicate` and `extend`. I don't know that we need it since our spec only mentions `extend` and not `duplicate`.

FWIW, I've found defining `duplicate` to be so much easier than defining `extend` for most data types. Maybe that's something to consider. Of course, we then lose the ability to derive `map` for a `Comonad` (similar to how it was when `Traversable` required `sequence` and not `traverse`). But, we're not even suggesting that right now. Seems like this PR is raising more issues than originally intended :smile:.

joneshf:
Maybe a worthwhile law is the second one here under `duplicated`: https://hackage.haskell.org/package/semigroupoids-5.1/docs/Data-Functor-Extend.html#v:duplicated

```haskell
fmap (fmap f) . extend id = extend id . fmap f
```

Maybe that's just a free theorem?

safareli:
Don't know but nor in `HS.semigroupoids.Extend`, nor in `HS.comonad.Comonad` and nor in `PS.control.Extend` we have the law. all of them have just `extended f . extended g = extended (f . extended g)`  for case when `extend` is defined only. looks like it's not needed, so maybe we should remove it?

or we can just move the `map` law into Extend and create an issue to revisit the extend/comonad laws. @davidchambers in this case `/laws` should be updated too.

davidchambers:
Could someone propose appropriate names for the Extend and Comonad laws? When we moved a law from Comonad to Extend we ended up with two Extend laws *both* described as "associativity".

safareli:
That law is not even in [cats CoflatMap](https://github.com/typelevel/cats/blob/155f7f534993c30d6e757de990330ac796dad5da/laws/src/main/scala/cats/laws/CoflatMapLaws.scala).

I think we should just remove it, and if we in future have some way in the spec to say that for example Extend should define `extend` **or** `duplicate` then we should revisit laws for extend and comonad

jlmorgan:
Shouldn't everything that derives from `Functor` observe the same laws? Should you not be able to do:

```javascript
e.extend(a => a) is equivalent to e (identity)
e.extend(x => f(g(x))) is equivalent to e.extend(g).extend(f) (composition)
```

Personally, I think having the companion functions like `duplicate` and `join` provide better clarity as to _how_ both `Chain` and `Extend` are derived from `Functor`'s map. Providing how each of `Functor`'s descendants are implemented using `map` provides not only clarity, but a glimpse into intended uses of the morphisms.

joneshf:
> That law is not even in cats CoflatMap.

It's coherence: https://github.com/typelevel/cats/blob/155f7f534993c30d6e757de990330ac796dad5da/laws/src/main/scala/cats/laws/CoflatMapLaws.scala#L19-L20

> I think we should just remove it, and if we in future have some way in the spec to say that for example Extend should define extend or duplicate then we should revisit laws for extend and comonad

Sounds good to me.

joneshf:
@jlmorgan "identity" and "composition" as you've defined them there should not work, as you've shown. If they did work, then `extend` would just be `map`, and "associativity" would not work as you've defined it. In other words, there'd be no purpose for defining the `Extend` algebra; it would just be `Functor` again with a different name.

safareli:
@joneshf  the "coherence" uses `coflatten`.
@davidchambers 👍

joneshf:
I'm confused. What are we talking about?

safareli:
> > That law is not even in cats CoflatMap.
> 
> It's coherence: https://github.com/typelevel/cats/blob/155f7f534993c30d6e757de990330ac796dad5da/laws/src/main/scala/cats/laws/CoflatMapLaws.scala#L19-L20

What i'm saying is that the `coherence` law also depends on `coflatten` (duplicate) which we don't have here.

davidchambers:
Shall we merge this pull request, @joneshf?

SimonRichardson:
😂

Issue #217, opened 2017-01-02
Terminology improvement suggestion


dmitriz:
It seems that the same letters are used for different things if I see it right.

This is where it starts getting interesting:
https://github.com/fantasyland/fantasy-land#apply

Here `a` and `b` are the types:

```hs
ap :: Apply f => f a ~> f (a -> b) -> f b
```

But here `a` and `b` are values:

```js
a.ap(b)
```
>1. `b` must be an Apply of a function,

Further,  the `b` below corresponds to the value of type `f (a -> b)` above, not even of type `b` above, which is also present.

I've thought, using some different notation for the "lifted types" might help the reading.
Like `x` for a value of type `a`, `xy` for a value of type `f (a -> b)` and `y` for the result:

```js
y = x.ap(xy)
```

Also some quick examples may further help the understanding.
Like in this case I find the `Array` example a good illustration of
the difference between Apply and Functor:

```js
[' Sara', 'John  '].ap([v => v.trim(), v => v.length])
// => ['Sara', 'John', 5, 6] (the data wins, as in Haskell)
// or
// => ['Sara', 5, 'John', 6] (the function wins)
```

There is also further confusion with JS (and [Ramda's](http://ramdajs.com/docs/#apply)) `apply` method:

```js
// Below (a) is the tuple type, e.g. (1, 4, 5)
// ((a) -> b) ~>  [a] -> b
fn.apply(array)
```
which has arguments in the flipped order comparing to the current FL spec, 
and takes a plain function rather than a lifted one (i.e. of type `f a`).

On the other hand, [Ramda's lift](http://ramdajs.com/docs/#lift) is stating to satisfy the FL's Apply spec, but actually it takes a variadic "unlifted" function instead of a lifted unary function (i.e. `f(a -> b)`):

```js
// R.lift :: ((a) -> b) -> ([a]) -> [b]
```
Basically `R.lift` acts like Functor's `map` rather than an `ap`, if I see it right.

Let me know what you think.

SimonRichardson:
Hi @dmitriz thanks for the issue:

> It seems that the same letters are used for different things if I see it right.

This has been discussed before to some degree, I believe the out come was something on the lines of, we're happy to change things if it improves readability, but not for the sake of it! Also I think with the inclusion of the new type signatures it should be less of a pain and the types should talk for themselves.

> Also some quick examples may further help the understanding.

I'm not sure about this, maybe if we ever split the spec up into multiple documents then I would be more of a fan.

[...]

PR #218, opened 2017-01-03
Notation fix

dmitriz:
Some small notation improvements introducing some consistency rules:

- Types are denoted by letters at the beginning of alphabet : `a, b, c`

- Values are denoted by letters at the end of alphabet: `x, y, z`

- Unlifted values are denoted by `t, u, v`

- Lifted functions in Apply are called `xy` to distinguish from lifted values denoted by single letters like `x`.

(Partially) fixes https://github.com/fantasyland/fantasy-land/issues/217

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ A value which has a Setoid must provide an `equals` method. The
 
 ### Semigroup
 
-1. `a.concat(b).concat(c)` is equivalent to `a.concat(b.concat(c))` (associativity)
+1. `x.concat(y).concat(z)` is equivalent to `x.concat(y.concat(z))` (associativity)
 
 #### `concat` method
 
@@ -126,11 +126,11 @@ concat :: Semigroup a => a ~> a -> a
 A value which has a Semigroup must provide a `concat` method. The
 `concat` method takes one argument:
 
-    s.concat(b)
+    x.concat(y)
 
-1. `b` must be a value of the same Semigroup
+1. `x` and `y` must be values of the same Semigroup a
 
-    1. If `b` is not the same semigroup, behaviour of `concat` is
+    1. If `x` and `y` are not in the same semigroup, behaviour of `concat` is
        unspecified.
 
 2. `concat` must return a value of the same Semigroup.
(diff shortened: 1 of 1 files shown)


dmitriz:
In addition I would propose the standard mathematical convention to denote objects (aka types) and functors by capital letters. This is also consistent with the type (aka interface) and functor capitalised names e.g. `Maybe`, `Either`. 

## Further motivation

Functors are fundamentally different from functions (aka morphisms in Category theory). Functors map types to types, whereas functions map values to values. So it is highly beneficial for readability to make a clear notational distinction.

davidchambers:
We should be wary of trying to encode information in names. While these changes don't hurt, I don't think they're an improvement. I'm in favour of closing this pull request.

paldepind:
> We should be wary of trying to encode information in names.

Why? Encoding information in names seems very helpful to me. And it is practiced all the time in both math and programming.

davidchambers:
As soon as one gives functions and parameters descriptive names, readers will begin to make assumptions based on those names. If those assumptions are correct, fantastic! If those assumptions are *incorrect*, we'd have been better off sticking with meaningless names. :stuck_out_tongue_winking_eye:

I'm in favour of following existing conventions. I certainly find `f(x)` easier to parse than `x(f)`, though they're objectively equivalent. This pull request, though, introduces new conventions. The two-letter variable names, in particular, do not sit well with me.

paldepind:
> As soon as one gives functions and parameters descriptive names, readers will begin to make assumptions based on those names. If those assumptions are correct, fantastic! If those assumptions are incorrect, we'd have been better off sticking with meaningless names. :stuck_out_tongue_winking_eye:

Well. I personally think that well-chosen names are better than arbitrarily chosen name. Meaningful names are pretty much the norm so maybe users will mistake the arbitrary names for meaningful names and make incorrect assumptions about them anyway?

I do however agree about the two-letter names. What if just those where removed from the PR?

davidchambers:
> I do however agree about the two-letter names. What if just those where removed from the PR?

That would take me from :pensive: to :neutral_face:.

Issue #174, opened 2016-09-21
proposal: remove laws directory from repository


davidchambers:
Earlier today I created #170 and #172, but I now realize that many of the property tests are dysfunctional. To provide one more example, the test for semigroup associativity seeks to assert:

```
(a `concat` b) `concat` c = a `concat` (b `concat` c)
```

Since it uses the [same value](https://github.com/fantasyland/fantasy-land/blob/1.0.1/laws/semigroup.js#L15-L17) for `a`, `b`, and `c`, it actually asserts:

```
(a `concat` a) `concat` a = a `concat` (a `concat` a)
```

If `a` is `Identity('x')` and both sides evaluate to `Identity('xxx')`, we haven't proven the `fantasy-land/concat` method of our Identity type to be associative, have we?

The functions provided in `fantasy-land/laws` are not very useful. To _actually_ test many of these properties we can't use a universally applicable function such as `identity :: a -> a`. Instead, we must use functions specialized to the data types being tested. Choosing these functions requires a great deal of care, and _this_ is the thing which makes testing these properties challenging, not copying a line from the spec and adding the `fantasy-land/` prefixes where appropriate.

SimonRichardson:
fyi: So the issue arose when removing the laws from their original tests and making the laws more generalised. This obviously isn't ideal and not fully baked, but I do think we should provide a solution to how do you test the spec. This might not be the way, but we should have an avenue for discussing it.

Easy: Provide more arguments to allow to test more values (like the original tests when they where in fantasy-check).

davidchambers:
> Easy: Provide more arguments to allow to test more values (like the original tests when they where in fantasy-check).

Let's consider a concrete example: asserting associativity of `Maybe#fantasy-land/concat`.

Without `fantasy-land/laws`:

```javascript
const fl = require('fantasy-land');
const jsc = require('jsverify');

const equals = ...;

//    JustArb :: Arbitrary a -> Arbitrary (Maybe a)
const JustArb = arb => arb.smap(S.Just, m => m.value, R.toString);

//    MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
const MaybeArb = arb => jsc.oneof(JustArb(arb), jsc.constant(S.Nothing));

describe('Maybe', () => {

  describe('Semigroup', () => {

    it('satisfies associativity', () => {
[...]

PR #220, opened 2017-01-14
have traverse take a type representative rather than an "of" function

### LINKED ISSUE #207: use `TypeRep f` in traverse instead of `a -> f a`
As continuation of #206 I think we should change the signature of traverse
```diff
-traverse :: Applicative f, Traversable t => t a ~> (a -> f b, c -> f c) -> f (t b)
+traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
```

`sanctuary-type-classes` TypeReps are first arguments and i think it's fine, but if someone things that `TypeRep f` should be last argument let's discuss that too.

**This would be a breaking change so we might hold it for some time.**

davidchambers:
Closes #207

I intend to merge sanctuary-js/sanctuary-type-classes#28 regardless, but it would be nice to make this improvement here too. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,13 +340,15 @@ method takes two arguments:
 A value that implements the Traversable specification must also
 implement the [Functor](#functor) and [Foldable](#foldable) specifications.
 
-1. `t(u.traverse(x => x, F.of))` is equivalent to `u.traverse(t, G.of)`
-for any `t` such that `t(a).map(f)` is equivalent to `t(a.map(f))` (naturality)
+1. `t(u.traverse(F, x => x))` is equivalent to `u.traverse(G, t)` for any
+   `t` such that `t(a).map(f)` is equivalent to `t(a.map(f))` (naturality)
 
-2. `u.traverse(F.of, F.of)` is equivalent to `F.of(u)` for any Applicative `F` (identity)
+2. `u.traverse(F, F.of)` is equivalent to `F.of(u)` for any Applicative `F`
+   (identity)
 
-3. `u.traverse(x => new Compose(x), Compose.of)` is equivalent to
-   `new Compose(u.traverse(x => x, F.of).map(x => x.traverse(x => x, G.of)))` for `Compose` defined below and any Applicatives `F` and `G` (composition)
+3. `u.traverse(Compose, x => new Compose(x))` is equivalent to
+   `new Compose(u.traverse(F, x => x).map(x => x.traverse(G, x => x)))` for
+   `Compose` defined below and any Applicatives `F` and `G` (composition)
 
 ```js
 var Compose = function(c) {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -369,22 +371,24 @@ Compose.prototype.map = function(f) {
...
 ```hs
-traverse :: Applicative f, Traversable t => t a ~> (a -> f b, c -> f c) -> f (t b)
+traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
 ```
 
 A value which has a Traversable must provide a `traverse` method. The `traverse`
 method takes two arguments:
 
-    u.traverse(f, of)
+    u.traverse(A, f)
```
evilsoft: So HAPPY right now that the order was finally switched. 🎈 🍰 🎉 🌈 🦄

davidchambers: I didn't like the old order, but I couldn't justify a breaking change to scratch my itch. But if we're making a breaking change anyway we can flip the parameters while we're at it. :)

Out of interest, why do you prefer the new order?

evilsoft: Well for me, the TR is the more stable of the two parameters. It defines a classification if you will.

I could have a whole family of `a -> f b` functions, but they all tie to one `f`.

So using some point-free `traverse`, it just seems natural to fix that parameter and get a function back for that fixed Applicative. If that makes sense?

davidchambers: I'm with you :100: per cent. :)

laws/traversable.js
```diff
@@ -8,33 +8,35 @@ const {of, traverse, map} = require('..');
...
+  const a = identity(t(x)[traverse](T, y => y));
+  const b = t(x)[traverse](T, identity);
   return eq(a, b);
 };
 
-const identityʹ = t => eq => x => {
+const identityʹ = T => t => eq => x => {
   const u = [x];
 
-  const a = u[traverse](Id[of], Id[of]);
```
safareli: I think in this old version, it should have been:
```js
const a = u[traverse](Id[of], Id[of]);
const b = t(u);
```

and because of that, new version should be like this:
```js
const identityʹ = T => eq => x => {
  const u = [x];
  const a = u[traverse](T[of], T);
  const b = T[of](u);
  return eq(a, b);
}
```

and on caller side:
```js
test(traversable.identity(Id)(equality))
```

internal/patch.js
```diff
@@ -7,19 +7,23 @@ module.exports = () => {
   Array.prototype[fl.equals] = function(y) {
     return this.length === y.length && this.join('') === y.join('');
   };
-  Array.prototype[fl.map] = Array.prototype.map;
+  Array.prototype[fl.map] = function(f) {
+    return this.map(x => f(x));
+  };
```
davidchambers: In addition to mutating globally defined prototypes and featuring bugs such as the one above, this file contains a hack for `Array#fantasy-land/equals`, convincing me that the laws do not belong in this repository (#174). We really need [sanctuary-type-classes][1] to avoid monkey patching.

Sanctuary's property-based tests are really nice. See, for example, [__Functor.js__][2]. I'd like to move these tests to __fantasyland/fantasy-laws__ and remove the __laws__ directory from this repository. :)


[1]: https://github.com/sanctuary-js/sanctuary-type-classes
[2]: https://github.com/sanctuary-js/sanctuary/blob/master/test/internal/laws/Functor.js

SimonRichardson: @davidchambers DO IT.

davidchambers: Will do. :)

Sanctuary is missing some of the laws, as we've only defined tests for those applicable to Maybe and/or Either. sanctuary-js/sanctuary#323 adds tests for Alt, Plus, and Alternative. Once that pull request has been merged I'll create __fantasyland/fantasy-laws__ and submit an initial pull request.

SimonRichardson: Thanks

Issue #222, opened 2017-01-29
Link to Hindley-Milner notation explanation


justin-calleja:
I'm trying to understand some of the method definitions… specifically, what does "~>" mean in the Hindley-Milner notation (is it even Hindley-Milner notation? Got this from Brian Lonsdorf's https://drboolean.gitbooks.io/mostly-adequate-guide/content/ch7.html).

E.g.

concat :: Semigroup a => a ~> a -> a

Here's what I understand:

AFAIK, this notations usually describes functions. The only thing I started out knowing about this notation before doing some research was that it's input1 -> input2 -> ... -> inputN -> output.

In any case, `concat` is not a function, it's a method - there's an example in the doc: `s.concat(b)`.
So `concat` is a method on an object of "type" Semigroup (i.e. this object implements Semigroup's interface).
In `concat`'s definition, 'a' is an object which is of type Semigroup because of "Semigroup a =>" part. This is one of the things I didn't know upon first landing on the Fantasy Land doc page and later found out from Brian's book. I landed on the Fantasy Land doc page many times without knowing this (just mentioning in case anyone wants to know some of the issues beginners like myself may have on first encounters with fantasy-land).

Then the definition continues with: "a ~> a -> a"

But `concat` only actually takes one input, from the e.g. `s.concat(b)`. So a reasonable explanation for "~>" is that it means something like "yes LHS is an argument but via dot notation, not passing it in as an argument" - or something like that.

In any case, could someone please add a link in the README to an explanation of this notation for people who are not familiar with it (or Haskell, or Category theory)?

Many thanks

SimonRichardson:
This might help https://pursuit.purescript.org/packages/purescript-prelude/2.1.0/docs/Data.NaturalTransformation

safareli:
@SimonRichardson how Natural Transformation could help? 
I mean in haskell `~>` is natural transformation, but we use it here to describe that left side is type of `this` and right side is `arguments`, as i understand it has no connection with Natural Transformations. (https://github.com/fantasyland/fantasy-land/pull/147#issue-170054090)

SimonRichardson:
Does it not :embarrassed: oops! I've been using purescript too much recently.

SimonRichardson:
If I've forgotten about that, then we should definitely have a bit about it stating the fact!

davidchambers:
<https://sanctuary.js.org/#types> provides an explanation, @justin-calleja. We should link to this section or include something similar in the FL readme itself.

[...]

PR #223, opened 2017-01-31
Added a section on "Type signature notation" in README.md

justin-calleja:
Proposed fix for issue #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,73 @@ have dependencies on other algebras which must be implemented.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## Type signature notation
+
+The type signature notation used in this document is described below:<sup
+id="sanctuary-types-return">[1](#sanctuary-types)</sup>
+
+* `::` _"is a member of"._
+    - `e :: t` can be read as: "the expression `e` is a member of type `t`".
+    - `true :: Boolean` - "`true` is a member of type `Boolean`".
+    - `42 :: Integer, Number` - "`42` is a member of type `Integer and
+      Number`".
+* _New types can be created via type constructors._
+    - Type constructors can take zero or more type arguments.
+    - `Array` is a type constructor which takes one type argument.
+    - `Array String` is the type of all arrays of strings. Each of the
+      following has type `Array String`: `[]`, `['foo', 'bar', 'baz']`.
+    - `Array (Array String)` is the type of all arrays of arrays of strings.
+      Each of the following has type `Array (Array String)`: `[]`, `[ [], []
+      ]`, `[ [], ['foo'], ['bar', 'baz'] ]`.
+* _Lowercase letters stand for type variables._
+    - Type variables can take any type unless they have been restricted by
+      means of type constraints (see fat arrow below).
+* `->` (arrow) _Function type constructor._
+    - `->` is an _infix_ type construcor that takes two type arguments and
+      gives the type of functions between these two type arguments.
+    - `->`'s input type can be a grouping of types to create the type of a
+      function which accepts more than one argument (i.e. has arity greater
+      than one). When used in this way, the syntax is: `(<input-types>) ->
+      <output-type>`, where <input-types> comprises two or more comma–space
+      (`, `)-separated type representations.
+    - `String -> Array String` is a type satisfied by functions which take a
+      `String` and return an `Array String`.
+    - `String -> Array String -> Array String` is a type satisfied by functions
+      which take a `String` and return a function which takes an `Array String`
+      and returns an `Array String`.
+    - `(String, Array String) -> Array String` is a type satisfied by functions
+      which take a `String` and an `Array String` as arguments and return an
+      `Array String`.
+* `~>` (squiggly arrow) _Method type constructor._
+    - When a function is a property of an Object, it is called a method. All
+      methods have an implicit parameter type - the type of which they are a
+      property.
+    - `a ~> a -> a` is a type satisfied by methods on Objects of type `a` which
+      take a type `a` as an argument and return a value of type `a`.
+* `=>` (fat arrow) _Expresses constraints on type variables._
+    - In `a ~> a -> a` (see squiggly arrow above), `a` can be of any type.
+      `Semigroup a => a ~> a -> a` adds a constraint such that the type `a`
+      must now satisfy the `Semigroup` typeclass. To satisfy a typeclass means
+      to implement any function/method defined by that typeclass. In
+      addition, any implementation of such function/method must also satisfy
+      any laws defined by a given typeclass for the function/method in
+      question.
+
+For example:
+
+```
+traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
+'------'    '--------------------------'    '-'    '-------------------'    '-----'
+ '           '                               '      '                        '
+ '           ' - type constraints            '      ' - argument types       ' - return type
+ '                                           '
+ '- method name                              ' - method target type
+```
+
+- - -
+1. <a name="sanctuary-types"></a>See the [Types](https://sanctuary.js.org/#types)
+   section in Sanctuary's docs for more info. [↩](#sanctuary-types-return)
+
 ## Prefixed method names
 
 In order for a data type to be compatible with Fantasy Land, its values must


davidchambers:
Thanks, @justin-calleja! We should certainly include something like this.

README.md
```diff
@@ -50,6 +50,22 @@ have dependencies on other algebras which must be implemented.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## Type signature notation
+
+The type signature notation used in this document is described [here](https://sanctuary.js.org/#types). In summary:
+
+1. `->` (arrow) expresses a function's explicit parameter and return value types. `String` -> `[String]` -> `[String]` is a function which takes a `String` and an `[String]` and returns an `[String]`.
```
davidchambers: Haskell's special `[]` type constructor is not something we should emulate, in my view. In Sanctuary we've switched to using `Array :: Type -> Type` (as PureScript did quite a while ago), as it's more consistent:

```purescript
maybeToArray :: Maybe a -> [a]
arrayToMaybe :: [a] -> Maybe a
```

```purescript
maybeToArray :: Maybe a -> Array a
arrayToMaybe :: Array a -> Maybe a
```

Before introducing `->` via `String -> Array String -> Array String` it would be good to introduce type constructors so readers know what `Array String` represents.

README.md
```diff
@@ -50,6 +50,22 @@ have dependencies on other algebras which must be implemented.
...
+2. `~>` (squiggly arrow) expresses a function's implicit parameter type. When a function is a property of a type, it is called a method. All methods have an implicit parameter type - the type of which they are a property of.
+3. `=>` (fat arrow) expresses constraints on type variables. For example, `a -> b` is a function which takes an argument and returns a result. The argument can be any type. The result can be any type (including the same type as the argument). All this applies to the function described by `Functor f => f a -> f b` but we now require the type `f` to satisfy the Functor specification.
+
+i.e.
+
+```
+sequence :: Apply f, Traversable t => t (f a) ~> (b -> f b) -> f (t a)
+'------'    '--------------------'    '-----'    '-------------------'
+   '- method name     '- constraints     '- method target  '- method arguments
```
davidchambers: We now require `traverse` rather than `sequence`. It would be good to update this. Also, the section labelled "method arguments" is actually argument type and return type.

safareli:
```
traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
'------'    '--------------------------'    '-'    '-------------------'    '-----'
 |           |                               |      |                        ^ result
 |           |                               |      ^ method arguments.
 |           |                               ^ method target (`this`)
 |           ^ constraints on type variables
 ^ method name
```

justin-calleja: Note: I changed the example to `map`'s definition, because in `traverse`'s:

```
traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
```

I don't understand what `TypeRep` is. I know there's this section in the README (but I still don't get it): 

```
## Type representatives

Certain behaviours are defined from the perspective of a member of a type.
Other behaviours do not require a member. Thus certain algebras require a
type to provide a value-level representative (with certain properties). The
Identity type, for example, could provide `Id` as its type representative:
`Id :: TypeRep Identity`.

If a type provides a type representative, each member of the type must have
a `constructor` property which is a reference to the type representative.
```

Maybe someone can add an example?

safareli: I think traverse is good example as it contains multiple constraints and multiple arguments

justin-calleja: @safareli , fair enough - I'll change it later and open another issue for an example of TypeRep.

safareli: @justin-calleja this https://github.com/fantasyland/fantasy-land/pull/180#discussion_r81567422 thread might be relevant for understanding TypeRep.

README.md
```diff
@@ -50,6 +50,28 @@ have dependencies on other algebras which must be implemented.
...
+2. Lowercase letters stand for type variables. Type variables can take any type unless they have been restricted by means of type constraints (see fat arrow below).
+3. `->` (arrow) expresses a function's explicit parameter and return value types. `String` -> `Array String` -> `Array String` is a function which takes a `String` and an `Array String` and returns an `Array String`.
+4. `~>` (squiggly arrow) expresses a function's implicit parameter type. When a function is a property of a type, it is called a method. All methods have an implicit parameter type - the type of which they are a property of.
+5. `=>` (fat arrow) expresses constraints on type variables. For example, `a -> b` is a function which takes an argument and returns a result. The argument can be any type. The result can be any type (including the same type as the argument). All this applies to the function described by `Functor f => f a -> f b` but we now require the type `f` to satisfy the Functor specification.
+
+i.e.
```
davidchambers: Let's change this line:

```diff
-i.e.
+For example:
```

README.md
```diff
@@ -50,6 +50,28 @@ have dependencies on other algebras which must be implemented.
...
+
+```
+traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
+'------'    '--------------------------'    '-'    '-------------------'    '-----'
+ '           '                               '      '                        ' - return type
+ '           '                               '      ' - argument type
+ '           '                               ' - method target type
+ '           ' - type constraint
+ '- method name
```
davidchambers: I think this would look better if `type constraint` and `method name` were aligned with `argument type` and `method target type` respectively. In other words, let's move these up two lines. :)

README.md
```diff
@@ -50,6 +50,28 @@ have dependencies on other algebras which must be implemented.
...
+2. Lowercase letters stand for type variables. Type variables can take any type unless they have been restricted by means of type constraints (see fat arrow below).
+3. `->` (arrow) expresses a function's explicit parameter and return value types. `String` -> `Array String` -> `Array String` is a function which takes a `String` and an `Array String` and returns an `Array String`.
+4. `~>` (squiggly arrow) expresses a function's implicit parameter type. When a function is a property of a type, it is called a method. All methods have an implicit parameter type - the type of which they are a property of.
+5. `=>` (fat arrow) expresses constraints on type variables. For example, `a -> b` is a function which takes an argument and returns a result. The argument can be any type. The result can be any type (including the same type as the argument). All this applies to the function described by `Functor f => f a -> f b` but we now require the type `f` to satisfy the Functor specification.
```
davidchambers: Please wrap lines at 79 characters. Long lines make line comments less precise.

README.md
```diff
@@ -50,6 +50,28 @@ have dependencies on other algebras which must be implemented.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## Type signature notation
+
+The type signature notation used in this document is described [here](https://sanctuary.js.org/#types). In summary:
+
+1. New types can be created via type constructors. Type constructors take one or more type arguments. For example, the type "array of strings" is written as `Array String`, and the type constructor `Array` can be defined as `Array :: a -> a`.
```
davidchambers: The type of `Array` is not `a -> a` but `Type -> Type`.

README.md
```diff
@@ -50,6 +50,28 @@ have dependencies on other algebras which must be implemented.
...
+2. Lowercase letters stand for type variables. Type variables can take any type unless they have been restricted by means of type constraints (see fat arrow below).
+3. `->` (arrow) expresses a function's explicit parameter and return value types. `String` -> `Array String` -> `Array String` is a function which takes a `String` and an `Array String` and returns an `Array String`.
+4. `~>` (squiggly arrow) expresses a function's implicit parameter type. When a function is a property of a type, it is called a method. All methods have an implicit parameter type - the type of which they are a property of.
```
davidchambers: We could drop the last word of the last sentence to avoid the double "of".

README.md
davidchambers: I think `Semigroup a => a -> a -> a` is a clearer example of a type-class constraint.

rstegg: I love this suggestion, and if I may make a suggestion?  This is looking like a dictionary, so perhaps keeping concise and declarative definitions to a single sentence, for instance - "`=>`: expresses constraints on type variables." to make a reference point if the reader needs it [at least, IMO that is how the rest of the specs read] (and possibly adds illustration as to what the signatures are used for)

README.md
justin-calleja: @davidchambers Is this good:

```
map :: Functor f => f a ~> (a -> b) -> f b
'-'    '-------'    '-'    '------'    '-'
 '      '            '      '           '
 '- method name      ' - method target type
        '                   '           '
        ' - type constraint ' - argument type
                                        '
                                        ' - return type
```

Or should "return type" be level 1? Also, what about space between `type constraint ` and `argument type`? I kept it so small for consistency (all down arrows start from first "-" in the thing they underline).

README.md
justin-calleja: @davidchambers Aren't lowercase letters used for type variables? What is `Type`? I would imagine you mean a type variable by it - but then it starts with capital case (so it could mean some defined type called "Type").

README.md
justin-calleja: @davidchambers , I'm wrapping only this PR's content at 79 characters. Do you want another PR with the whole README.md wrapped at 79 chars?

justin-calleja:
btw, just noticed I'm using past tense in commit messages (whereas other commit messages use imperative tense).

I'm not sure how to change this now that I've pushed to my repo. I could delete the repo and add all these changes as one commit with imperative tense in another fork. I also have no problem with anyone just taking the changes and committing properly with their username - seriously, I just want the info on the README.md.

README.md
justin-calleja: @rstegg true - for a "summary" it is getting a bit long. Basically, I want to target ppl who know next to nothing re the type signature notation. There is a reference to `Sanctuary`'s docs - but I'd rather not have someone stay fishing for the relevant parts of the notation and just have it listed here. 

I didn't see this comment originally as I would have made the changes with the last commit. I suggest the following:

1. `=>` _expresses constraints on type variables._
    More info / example for total newcomers here.
2. `->` _expresses a function's explicit parameter and return value types._
    More info / example for total newcomers here.

etc… (i.e. italic short description followed by new line + example / longer description)

There is now also the issue of tense in commit messages. @davidchambers please advise on what to do re that. Should I scrap the repo I'm making this PR with and create a new one with the current changes and @rstegg 's feedback?

davidchambers:
> I'm not sure how to change this now that I've pushed to my repo. I could delete the repo and add all these changes as one commit with imperative tense in another fork.

No need for such drastic action! :stuck_out_tongue_winking_eye:

The nice thing about working on a feature branch is that one's free to `--force` push:

```console
$ git push origin issue-222 --force
```

Editing the last commit is quite easy. Simply `git add` some changes then run:

```console
$ git commit --amend
```

To edit commit `abcd123` one would run:

```console
$ git rebase abcd123^ --interactive
```

Feel free to send me a message on Gitter if you have Git-related questions, @justin-calleja.

README.md
davidchambers: There really is a `Type` type: sanctuary-js/sanctuary-def#125.

If we're talking the [`$.Array`][1] type constructor, it is not polymorphic – it must be applied to a `Type` value such as [`$.Integer`][2]. `Array :: a -> a` would be an oddly named identity function. ;)


[1]: https://github.com/sanctuary-js/sanctuary-def#Array
[2]: https://github.com/sanctuary-js/sanctuary-def#Integer

justin-calleja:
@davidchambers thanks! If only the force were always a flag away :laughing:

Squashed everything to a single commit - seems to have kept previous conversation history here in github (OS newbie)

justin-calleja:
@safareli I just noticed I never updated the example to use `traverse`. I think it's because I never got what `TypeRep` is and I was going to do that after (in case I could add some more info).

In any case, updated it now. @davidchambers note that the example is over 79 chars long (otherwise it would break the format of the e.g. - some other lines in the README are also over 79 chars long).

README.md
justin-calleja: Thanks David. I'm still grokking this - I think I need to read up a bit more on type constructors (looking at Sanctuary).

I do get this though:

> Array :: a -> a would be an oddly named identity function

so `Array :: a -> a` is wrong - but I'm still looking into why `Array :: Type -> Type` is right.

davidchambers: These `ghci` (Haskell) examples may be instructive:

```haskell
> :kind String
String :: *

> :kind Maybe
Maybe :: * -> *

> :kind Maybe String
Maybe String :: *
```

One can think of `*` as the type of types. The type of `String` is `*` – `String` is a type. The type of `Maybe` is `* -> *` – `Maybe` is a type *constructor*.

README.md
```diff
@@ -50,6 +50,42 @@ have dependencies on other algebras which must be implemented.
...
+For example:
+
+```
+traverse :: Applicative f, Traversable t => t a ~> (TypeRep f, a -> f b) -> f (t b)
+'------'    '--------------------------'    '-'    '-------------------'    '-----'
+ '           '                               '      '                        '
+ '           ' - type constraints            '      ' - argument types       ' - return type
+ '                                           '
+ '- method name                              ' - method target type
+```
```
davidchambers: The updated formatting looks good!

README.md
```diff
@@ -50,6 +50,42 @@ have dependencies on other algebras which must be implemented.
...
 
+## Type signature notation
+
+The type signature notation used in this document is described
+[here](https://sanctuary.js.org/#types). In summary:
+
+1. _New types can be created via type constructors._
+    - Type constructors take one or more type arguments. For example, the type
+    "array of strings" is written as `Array String`, and the type constructor
+    `Array` can be defined as `Array :: Type -> Type`.
```
davidchambers: I think this point should come last, or be removed. An explanation of the notation should start with the simplest possible example and build from there. Type constructors should not be introduced until the groundwork has been laid.

I suggest something like this:

```haskell
-- simple example
true :: Boolean
```

```haskell
-- show that a value can be a member of multiple types
42 :: Integer, Number
```

```haskell
-- introduce type constructors
["foo", "bar", "baz"] :: Array String
```

```haskell
-- introduce the (->) type constructor
Math.abs :: Number -> Number
```

README.md
```diff
@@ -50,6 +50,42 @@ have dependencies on other algebras which must be implemented.
...
+    - `String` -> `Array String` -> `Array String` is a function which
+    takes a `String` and an `Array String` and returns an `Array String`.
+4. `~>` (squiggly arrow) _expresses a function's implicit parameter type._
+    - When a function is a property of a type, it is called a method. All methods
+    have an implicit parameter type - the type of which they are a property.
+5. `=>` (fat arrow) _expresses constraints on type variables._
+    - For example, `a ~> a -> a` is a method on some type `a`, which takes an
+    argument and returns a result, both of which must be of the same type `a`.
+    All this applies to the method described by `Semigroup a => a ~> a -> a` but
+    we now require the type `a` to also satisfy the Semigroup specification.
```
davidchambers: I believe this list should be a `<ul>` rather than an `<ol>`.

README.md
justin-calleja: @davidchambers Ok it seems to be the case that "type constructors" wrap their type arguments (which can be variables or not (a v.s. Type)).

On interpreting this: `Array :: Type -> Type`, I was thinking "Array, given a Type, will give you back a type of Type". But I was expecting it to give back an "Array of Type" ... not just Type ... and I couldn't resolve this.

But if "by default", any type returned by a type constructor is wrapped by that type constructor, then it makes sense.

```
> :kind Maybe
Maybe :: * -> *
```

`Maybe` given a Type gives back a `Maybe Type`. `Array`, given a type gives back an `Array Type` not just a `Type`.

Thanks!

README.md
```diff
@@ -50,6 +50,42 @@ have dependencies on other algebras which must be implemented.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## Type signature notation
+
+The type signature notation used in this document is described
+[here](https://sanctuary.js.org/#types). In summary:
```
alex-deas: Is linking to the Sanctuary type defs adequate. If Fantasy land is intending to be a specification for API's, should it not define type defs itself rather than deferring to other libraries? Probably beyond the scope of this change, but curious to know what other people's thoughts are for defining a type syntax as part of FL

justin-calleja: @alex-deas - personally, I'd link to anything I might think useful to people without prior knowledge of Haskell. The type signature notation is based on Haskell's - with (AFAIK) some extensions since we're annotating JS code (like the `~>` for "method type constructor").

I am currently re-writing this PR and think I'll make the link to Sanctuary less prominent (i.e. footnote - "for more, see Link")

README.md
davidchambers: > `Maybe` given a Type gives back a `Maybe Type`. `Array`, given a type gives back an `Array Type` not just a `Type`.

This is incorrect. An array of types looks like this:

```purescript
[$.Boolean, $.Number, $.String] :: Array Type
```

`$.Array($.Integer)`, on the other hand, represents a single type—`Array Integer`—rather than an array of types:

```purescript
$.Array :: Type -> Type
$.Array($.Integer) :: Type
```

It's important to keep in mind that the type of `$.Array($.Integer)` is *not* `Array Integer`. Thinking about the types of types is especially confusing in a language without types, where types—and the types of types—can only exist at the value level.

README.md
justin-calleja: @davidchambers I've re-written the section based on this info and some research: https://github.com/justin-calleja/fantasy-land/tree/issue-222#type-signature-notation

README.md
justin-calleja: @davidchambers I've updated [this section in the README.md](https://github.com/justin-calleja/fantasy-land/tree/issue-222#type-signature-notation). Basically, my current understanding is there so feel free to slice and dice it as necessary 😄 

I think I get it though. A lot hinges on properly understanding `Type`. I have defined it as the type of a "type expression" in the README.md (based on some looking around). Any thoughts on this?

davidchambers: > A lot hinges on properly understanding `Type`.

I think introducing a `Type` type actually makes things less clear. Another option is to provide a few examples to encourage an intuitive understanding rather than a theoretical one:

>   - `Array String` is the type comprising every array of strings. Examples:
>
>       - `[]`
>       - `['foo', 'bar', 'baz']`
>
>   - `Array (Array String)` is the type comprising every array of array of strings. Examples:
>
>       - `[]`
>       - `[[], [], [], []]`
>       - `[[], ['foo'], ['bar', 'baz']]`

README.md
```diff
@@ -50,6 +50,83 @@ have dependencies on other algebras which must be implemented.
     - Two promises are equivalent when they yield equivalent values.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
+## Type signature notation
+
+The type signature notation used in this document is described below:<sup
+id="sanctuary-types-return">[1](#sanctuary-types)</sup>
+
+* `::` _"has type of"._
```
davidchambers: I believe "is a member of" is preferable to "has type of". The latter implies that a value's type is a property of the value. It's better to encourage readers to think of types as sets of values. Just as the values in the set `{1,2,3}` are oblivious to their membership of the set, `true` and `false` are oblivious to their membership of `{true,false}` (which we refer to as `Boolean`).

Using "has type of" also encourages readers to think in terms of `typeof` and `instanceof`, which—ignoring inheritance—encourage us to think of every value as belonging in exactly one set of values.

README.md
```diff
@@ -50,6 +50,83 @@ have dependencies on other algebras which must be implemented.
...
+'------'    '--------------------------'    '-'    '-------------------'    '-----'
+ '           '                               '      '                        '
+ '           ' - type constraints            '      ' - argument types       ' - return type
+ '                                           '
+ '- method name                              ' - method target type
+```
+
+- - -
+1. <a name="sanctuary-types"></a>See the [Types](https://sanctuary.js.org/#types)
+   section in `Sanctuary.js`'s docs for more info. [↩](#sanctuary-types-return)
```
davidchambers: Sanctuary should be written without a file extension. :stuck_out_tongue_winking_eye:

README.md
justin-calleja: @davidchambers Ok, so if I’m understanding you correctly, when explaining “type constructors”, I’ll remove the actual definitions and use examples only instead.

This means we can also remove the part “the value of an expression may be a type” - since the whole point of introducing it was because I needed it to use actual type definitions in the “type constructor” part.

README.md
justin-calleja: hehe sure - bad habits ^^;

README.md
justin-calleja: Here's what I meant by my prev comment: https://github.com/justin-calleja/fantasy-land/commit/97da83cd7707dbb2a8d7984f2a3083fc19f93923

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
 
+## Type signature notation
+
+The type signature notation used in this document is described below:<sup
+id="sanctuary-types-return">[1](#sanctuary-types)</sup>
+
+* `::` _"is a member of"._
+    - `e :: t` can be read as: "the expression `e` is a member of type `t`".
+    - `true :: Boolean` - "`true` is a member of type `Boolean`".
+    - `42 :: Integer, Number` - "`42` is a member of type `Integer` or `Number`" (the
```
davidchambers: This should be _and_ rather than _or_.

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
+    - Type constructors can take zero or more type arguments.
+    - `Array` is a type constructor which takes one type argument.
+    - `Array String` is the type of all arrays of strings. Each of the
+      following has type `Array String`: `[]`, `['foo', 'bar', 'baz']`.
+    - `Array (Array String)` is the type of all arrays of arrays of strings.
+      Each of the following has type `Array (Array String)`: `[]`, `[ [], []
+      ]`, `[ [], ['foo'], ['bar', 'baz'] ]`.
+* _Lowercase letters stand for type variables._
+    - Type variables can take any type unless they have been restricted by means
+    of type constraints (see fat arrow below).
```
davidchambers: Nit: The indentation is off here.

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
+* `->` (arrow) _Function type constructor._
+    - `->` is an _infix_ type construcor that takes 2 type arguments and gives
+      the type of functions between these 2 type arguments.
+    - `->`'s input type can be a grouping of types to create the type of a
+      function which accepts more than one argument (i.e. has arity greater
+      than 1). When used in this way, the syntax is: `(<input-types>) ->
+      <output-type>`, where <input-types> comprises two or more comma–space
+      `(, )` -separated type representations.
```
davidchambers: I'm not sure how to interpret these parentheses.

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
+      `String` and return an `Array String`.
+    - `String -> Array String -> Array String` is a type satisfied by functions
+      which take a `String` and return a function which takes an `Array String`
+      and returns an `Array String`.
+    - `(String, Array String) -> Array String` is a type satisfied by functions
+      which take a `String` and an `Array String` as arguments and return an
+      `Array String`.
+* `~>` (squiggly arrow) _Method type constructor._
+    - When a function is a property of an Object, it is called a method. All methods
+    have an implicit parameter type - the type of which they are a property.
```
davidchambers: Nit: The indentation is off here.

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
+    - `a ~> a -> a` is a type satisfied by methods on Objects of type `a` which
+      take a type `a` as an argument and return a value of type `a`.
+* `=>` (fat arrow) _Expresses constraints on type variables._
+    - In `a ~> a -> a` (see squiggly arrow above), `a` can be of any type.
+      `Semigroup a => a ~> a -> a` adds a constraint such that the type `a`
+      must now satisfy the `Semigroup` typeclass. To satisfy a typeclass means
+      to implement any function / method defined by that typeclass. In
+      addition, any implementation of such function / method must also satisfy
+      any laws defined by a given typeclass for the function / method in
+      question.
```
davidchambers: I suggest dropping the spaces around the slash (`/`) characters.

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
+'------'    '--------------------------'    '-'    '-------------------'    '-----'
+ '           '                               '      '                        '
+ '           ' - type constraints            '      ' - argument types       ' - return type
+ '                                           '
+ '- method name                              ' - method target type
+```
+
+- - -
+1. <a name="sanctuary-types"></a>See the [Types](https://sanctuary.js.org/#types)
+   section in `Sanctuary`'s docs for more info. [↩](#sanctuary-types-return)
```
davidchambers: Nit: We needn't format _Sanctuary_ specially.

README.md
justin-calleja: @davidchambers I got that from https://sanctuary.js.org/#types

(search for "separated").

README.md
justin-calleja: @davidchambers I also removed this part as I think it's extra:

> (the value of the expression `42` is `42` and we claim that it qualifies for membership in both the `Integer` and `Number` types).

justin-calleja:
@davidchambers made the requested changes apart from the:

> `(, )` -separated

one. Not sure what to change there.

btw, I assume you want these rebased to a single commit right? I take it that that doesn't break any of the history here (comments in this chat).

README.md
davidchambers: There's a subtle but significant difference between `` `(, )` `` and `` (`, `) ``. ;)

davidchambers:
> I assume you want these rebased to a single commit right?

That would be nice, but we can leave that until the end.

> I take it that that doesn't break any of the history here (comments in this chat).

Correct. The only comments that become detached from a pull request when rebasing are commit comments, which means it's usually correct to make line comments on the overall diff rather than on individual commits.

README.md
justin-calleja: oh right ^^; nice catch!

README.md
```diff
@@ -50,6 +50,73 @@ have dependencies on other algebras which must be implemented.
...
+      following has type `Array String`: `[]`, `['foo', 'bar', 'baz']`.
+    - `Array (Array String)` is the type of all arrays of arrays of strings.
+      Each of the following has type `Array (Array String)`: `[]`, `[ [], []
+      ]`, `[ [], ['foo'], ['bar', 'baz'] ]`.
+* _Lowercase letters stand for type variables._
+    - Type variables can take any type unless they have been restricted by
+      means of type constraints (see fat arrow below).
+* `->` (arrow) _Function type constructor._
+    - `->` is an _infix_ type construcor that takes 2 type arguments and gives
+      the type of functions between these 2 type arguments.
```
davidchambers: In general I prefer to write single-digit numbers as words. I feel particularly strongly about this in a programming context, as it helps to differentiate the JavaScript value _`2`_ from the cardinal number _two_.

rjmk:
Thank you for sticking with this, Justin & David! It's going to be a very valuable patch :)

davidchambers:
@justin-calleja, please squash your commits and `git rebase master`. I'll then merge this. :)

justin-calleja:
@davidchambers thanks!

PR #226, opened 2017-02-12
Correct comonad value variable name

MichalZalecki:
Shouldn't that be `w` to be aligned with specifications below or am I missing something?


davidchambers:
LGTM

PR #229, opened 2017-03-13
Fix: map derivated from ap and of

shfshanyue:
it may be this from what i understand. for example 

``` js
const add = x => x + 1
[1, 2, 3, 4, 5].map(add)      // [2, 3, 4, 5, 6]
Array.of(add).ap([1, 2, 3, 4, 5]) // [2, 3, 4, 5, 6]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,7 +608,7 @@ to implement certain methods then derive the remaining methods. Derivations:
   - [`map`][] may be derived from [`ap`][] and [`of`][]:
 
     ```js
-    function(f) { return this.ap(this.of(f)); }
+    function(f) { return this.of(f).ap(this); }
     ```
 
   - [`map`][] may be derived from [`chain`][] and [`of`][]:


safareli:
[ap :: Apply f => f a ~> f (a -> b) -> f b](https://github.com/fantasyland/fantasy-land#ap-method)

davidchambers:
Do you know which source led you to this understanding, @shfshanyue? Perhaps ramda-fantasy or another library still targeting an old version of the Fantasy Land spec?

shfshanyue:
``` js
Array.prototype.ap = function (xs) {
  return this.reduce((acc, f) => acc.concat(xs.map(f)), [])
}
```

@davidchambers  it's it...

davidchambers:
Where did you find that snippet, @shfshanyue?

shfshanyue:
[applicative-functor](https://github.com/hemanth/functional-programming-jargon#applicative-functor)

thanks very much, @davidchambers

davidchambers:
That's in line with `ap` as it was defined prior to #145.

PR #230, opened 2017-03-16
fix: guard against null/undefined x

bumbleblym:
https://github.com/fantasyland/fantasy-land/blob/498e8236376df103101a09ac2dbffb4ac83ea9c5/internal/id.js#L11-L13

### DIFF
--- a/internal/func.js
+++ b/internal/func.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const fl = require('../');
-const equality = (x, y) => typeof x[fl.equals] === 'function' ? x[fl.equals](y) : x === y;
+const equality = (x, y) => x && typeof x[fl.equals] === 'function' ? x[fl.equals](y) : x === y;
 
 module.exports = {
   equality,


internal/func.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const fl = require('../');
-const equality = (x, y) => typeof x[fl.equals] === 'function' ? x[fl.equals](y) : x === y;
+const equality = (x, y) => x && typeof x[fl.equals] === 'function' ? x[fl.equals](y) : x === y;
```
Avaq: This new function will not strictly return a Boolean: `equality(null) === null`. Not sure if that's a problem, but if it is, I would suggest:

```js
const equality = (x, y) => Boolean(x) && ...
```

bumbleblym: the new function doesn't return `x`; isn't it the same as:

`const equality = (x, y) => (x && typeof x[fl.equals] === 'function') ? x[fl.equals](y) : x === y;`

[Operator precedence](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence)

Avaq: Ah. You're right. I didn't notice it's a ternary. My mistake.

bumbleblym: it happens 😆

Avaq:
Shall we close this in favor of moving towards fantasyland/fantasy-laws#1?

davidchambers:
Good idea, Aldwin.

PR #235, opened 2017-03-25
Add Ord spec

### LINKED ISSUE #233: Add spec for Ord
Motivated by https://github.com/sanctuary-js/sanctuary/issues/353 I think it makes sense to add `Ord`.

From [purescript-prelude](https://github.com/purescript/purescript-prelude/blob/b9c91d009fc035d142d3bbb8af52c55681b7b3db/src/Data/Ord.purs#L25-L34)

```purescript
-- | The `Ord` type class represents types which support comparisons with a
-- | _total order_.
-- |
-- | `Ord` instances should satisfy the laws of total orderings:
-- |
-- | - Reflexivity: `a <= a`
-- | - Antisymmetry: if `a <= b` and `b <= a` then `a = b`
-- | - Transitivity: if `a <= b` and `b <= c` then `a <= c`
class Eq a <= Ord a where
  compare :: a -> a -> Ordering
```

This _may_ require specifying `Ordering` as well (though I haven't seen much movement on spec'ing data types #185):

From [purescript-prelude](https://github.com/purescript/purescript-prelude/blob/b9c91d009fc035d142d3bbb8af52c55681b7b3db/src/Data/Ordering.purs#L7-L13)
```purescript
-- | The `Ordering` data type represents the three possible outcomes of
-- | comparing two values:
-- |
-- | `LT` - The first value is _less than_ the second.
-- | `GT` - The first value is _greater than_ the second.
-- | `EQ` - The first value is _equal to_ the second.
data Ordering = LT | GT | EQ
```

Specifying `Ord` would allow generalizing several comparison (`lt` and friends, `min` & `max`) as well as sorting functions (`sort`, `sortBy`).

gabejohnson:
Closes #233 

I intentionally left the definition of `Ordering` implicit (as is that of `Boolean` under `Setoid`) to spur discussion.

I'm looking for any suggestions on a better way to state the laws in both the README and the tests.


README.md
```diff
@@ -114,6 +114,33 @@ A value which has a Setoid must provide an `equals` method. The
 
 2. `equals` must return a boolean (`true` or `false`).
 
+### Ord
+
+A value that implements the Ord specification must also implement
+the [Setoid](#setoid) specification.
+
+1. `a.compare(a) <= 0` (reflexivity)
```
davidchambers: Why `<=` rather than `===`?

README.md
```diff
@@ -114,6 +114,33 @@ A value which has a Setoid must provide an `equals` method. The
...
+`compare` method takes one argument:
+
+     a.compare(b)
+
+1. `b` must be a value of the same Ord
+
+    1. If `b` is not the same Ord, behaviour of `compare` is
+       unspecified (returning NaN is recommended).
+
+2. `compare` must return a number (-1, 0 or 1).
```
davidchambers: I suggest backticks around `-1`, `0`, and `1` (and `NaN`). Also, an Oxford comma. ;)

README.md
gabejohnson: I was simply copying the laws from PureScript. No other reason. `===` works just as well and is probably clearer.

joneshf:
I'm not enamored with the idea of using `-1`, `0` and `1`. There are other formulations for orderings, posets and comparisons we can use that don't require this C-style obfuscation.

gabejohnson:
@joneshf what would you suggest? Strings? Symbols?

gabejohnson:
Okay here's another formulation:

1. `a.compare(a).equals([a, a])` (reflexivity)
2. If `a.compare(b).equals([b, a])` and `b.compare(a).equals([a, b])`, then `a.equals(b) === true` (antisymmetry)
3. If `a.compare(b).equals([b, a])` and `b.compare(c).equals([c, b])`, then `a.compare(c).equals([c, a])` (transitivity)

safareli:
dependencies.png should be regenerated

gabejohnson:
How about:
1. `a.covers(a) === false` (irreflexivity)
2. `a.covers(b) !== b.covers(a)` (asymmetry)
3. If `a.covers(b)` and `b.covers(c)`, then `a.covers(c)` (transitivity)

```js
// defined like operators
const lt = (x, y) => x.covers(y);
const gte = (x, y) => !x.covers(y);
const lte = (x, y) => x.covers(y) || x.equals(y);
const gt = (x, y) => !x.covers(y) && !x.equals(y);
const min = (x, y) => x.covers(y) ? x : y;
const max = (x, y) => x.covers(y) ? y : x;

const nativeCompare = (x, y) => x.covers(y) ? -1 : x.equals(y) ? 0 : 1;
```

gabejohnson:
Back to using `<=`:

1. `a.precedes(a) === true` (reflexivity);
2. If `a.precedes(b)` and `b.precedes(a)`, then `a.equals(b)` (antisymmetry)
3. If `a.precedes(b)` and `b.precedes(c)`, then `a.precedes(c)` (transitivity)

```js
// defined like operators
const lte = (x, y) => x.precedes(y);
const gt = (x, y) => !x.precedes(y);
const lt = (x, y) => x.precedes(y) && !x.equals(y);
const gte = (x, y) => !x.precedes(y) || x.equals(y);
const min = (x, y) => x.precedes(y) ? x : y;
const max = (x, y) => x.precedes(y) ? y : x;

const nativeCompare = (x, y) => x.equals(y) ? 0 : x.precedes(y) ? -1 : 1;
```

gabejohnson:
I just changed the formulation to use `precedes` instead of `compare`. Added a derivation for `equals` and regenerated dependencies.png.

README.md
```diff
@@ -605,6 +632,11 @@ The `profunctor` method takes two arguments:
 When creating data types which satisfy multiple algebras, authors may choose
 to implement certain methods then derive the remaining methods. Derivations:
 
+  - [`equals`][] may be derived from [`precedes`][]:
+    ```js
```
davidchambers: Nit: Empty line before `` ```js ``, please.

gabejohnson: Done

davidchambers:
I'll defer to @joneshf, @safareli, and others who are actually familiar with this stuff, but I will say that in its current state this pull request seems extremely elegant!

laws/ord.js
```diff
@@ -0,0 +1,45 @@
...
+  const g = t(x);
+
+  const a = f[precedes](g);
+  const b = g[precedes](f);
+  const c = g[equals](f);
+  const d = true;
+  return eq(a, d) && eq(b, d) && eq(c, d);
+};
+
+const transitivity = t => eq => x => {
```
safareli: Instead of passing `t`-s I think we should pass values on which `precedes` will be called.

safareli:
Small issue with laws otherwise LGTM. 
/cc @SimonRichardson @puffnfresh

laws/ord.js
gabejohnson: I defined `precedes` on `Id` which may have been wrong as it should just delegate to `value`'s `precedes`. Should I define it in `internal/func.js` instead? I could implement it for `String` and `Number` and delegate otherwise.

safareli: I think yes

gabejohnson:
Updated with changes to tests.

davidchambers:
Are you happy with this pull request in its current state, @joneshf? If so, let's merge it and release it as v3.2.0.

sadasant:
:shipit: ❤️

scott-christopher:
I would have assumed `precedes` meant `<` rather than `<=`. e.g. `(42).precedes(42) === true` doesn't sound right to me. How about `lte` as an alternative?

Besides naming, I welcome this addition.

davidchambers:
I like @scott-christopher's suggestion. Let's use `lte` rather than `precedes`.

joneshf:
I have a different qualm about naming. The name `Ord` comes from Haskell98. That typeclass is for total orderings. But here we're not requiring a total ordering. Although we don't really follow the typeclasses for other things (`Setoid` is a thing, `Monad` doesn't have methods, etc), I wonder how confusing this will be for someone coming from Haskell. So, I guess we should mention whether we want a partial ordering or a total ordering:

1. Rename to something like `Poset`, and mention that this is for partial orderings. Since we're actually specifying a poset here, might as well name it appropriately.
1. Keep the current name`Ord`, change the first law from reflexivity to totality, and mention that this is for total orderings.

I'm fine with either of those options. However, I think we should seriously reconsider it being unspecified currently.

davidchambers:
Thanks for the feedback, Hardy. :)

> However, I think we should seriously reconsider it being unspecified currently.

I don't understand this sentence. What is it that's unspecified?

joneshf:
I guess nothing, thinking about it. You can infer that it's for partial orderings and not total orderings based on the laws.

I think what I should have said was that we should be explicit, rather than saying that something was unspecified :blush:.

davidchambers:
I'm in favour of Hardy's second option as I believe the intention was to define the type class of totally ordered sets. Do you agree, @gabejohnson?

gabejohnson:
@davidchambers @joneshf that was an oversight on my part. I'll make it a total ordering unless you can think of a reason we wouldn't want to add that constraint.

gabejohnson:
@scott-christopher @davidchambers I was looking for a name that didn't suggest quantity and `precedes` is already used as a synonym for `lte` in some of the work I read, albeit inconsistently. I also thought it sounded nice paired with `equals`.
That said, since a total ordering can have a "least element" I suppose "less than" doesn't have to suggest quantity. I'll change it to `lte` unless anyone can think of a better name.

gabejohnson:
I've implemented all of the requested changes. Everything look good?

index.js
```diff
@@ -22,7 +22,8 @@
     extend: 'fantasy-land/extend',
     extract: 'fantasy-land/extract',
     bimap: 'fantasy-land/bimap',
-    promap: 'fantasy-land/promap'
+    promap: 'fantasy-land/promap',
+    lte: 'fantasy-land/lte'
```
davidchambers: Let's include this after `equals`.

gabejohnson:
Done.

README.md
```diff
@@ -605,6 +632,12 @@ The `profunctor` method takes two arguments:
 When creating data types which satisfy multiple algebras, authors may choose
 to implement certain methods then derive the remaining methods. Derivations:
 
+  - [`equals`][] may be derived from [`lte`][]:
+
+    ```js
+    function (b) { return this.lte(b) && b.lte(this); }
```
davidchambers: For consistency with the other derivations, let's drop the space after `function`.

I'd like to see the parameter named `other` rather than `b` to suggest that the method expects an argument of the same type.

README.md
```diff
@@ -700,6 +733,7 @@ be equivalent to that of the derivation (or derivations).
 [`extract`]: #extract-method
 [`map`]: #map-method
 [`of`]: #of-method
+[`lte`]: #lte-method
```
davidchambers: :abc:

laws/ord.js
```diff
@@ -0,0 +1,37 @@
...
+   3. If `a.lte(b)` and `b.lte(c)`, then `a.lte(c)` (transitivity)
+
+**/
+
+const totality = eq => f => g => {
+  const a = f[lte](g);
+  const b = g[lte](f);
+  const c = true;
+  return eq(a || b, c);
+};
```
davidchambers: Taking `eq` strikes me as unnecessary. I'd find this clearer:

```javascript
const totality = a => b => a[lte](b) || b[lte](a);
```

joneshf: I think all of the laws are like this—`equals` included. I can't remember why, but there was a reason we switched to this way of writing the laws.

joneshf: I.e., it seems fine to me to have these laws work similarly, and revisit whether that's appropriate in another PR. But, if you feel strongly, I won't fight it.

davidchambers: Given that I'm working on fantasyland/fantasy-laws#1, I'd very much like to know the reason. If it occurs to you, let me know. :)

Let's leave this as it is, Gabe. These functions will be removed from this repository in due course.

joneshf: Nice! I didn't know that existed.

gabejohnson:
I'll squash as soon as you sign off.

davidchambers:
Let's update the `height` of the `` for __dependencies.png__. The image's natural dimensions are now 1047 × 347. The width is fixed at 888; the height should be (888 / 1047) * 347 ≈ 294. ;)

gabejohnson:
Done

gabejohnson:
@davidchambers do you want me to squash it?

davidchambers:
Yes, it would be nice to squash the commits. :)

gabejohnson:
Squashed

gabejohnson:
Looks like I forgot to put the link to the new spec in the list [here](https://github.com/fantasyland/fantasy-land/pull/235/files#diff-04c6e90faac2675aa89e2176d2eec7d8R13)

davidchambers:
No worries. Would you mind opening another pull request? I'll then merge it and release v3.2.0.

PR #241, opened 2017-04-05
Add Enum spec

gabejohnson:
This a very rough first pass at defining a specification for `Enum` guided by Haskell's [`Enum`](https://hackage.haskell.org/package/base-4.9.1.0/docs/Prelude.html#t:Enum).

I'm open to splitting out the definitions of `fromEnum` and `toEnum` as well as renaming `fromEnum` and putting it on the instance instead.

The reason I restricted the values returned from `fromEnum` to being natural numbers is the same as outlined in the [Haskell Report](https://www.haskell.org/onlinereport/haskell2010/haskellch11.html#x18-18400011.2)

```js
const Ordering = Enum.deriveFromArray(['LT', 'EQ', 'GT']);
Ordering.fromEnum(Ordering.LT) // 0
Ordering.fromEnum(Ordering.EQ) // 1
Ordering.fromEnum(Ordering.GT) // 2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,8 @@ a `constructor` property which is a reference to the type representative.
 
 ### Enum
 
-1. `E.toEnum(E.fromEnum(E.a))` must be identical to `E.a`
-2. `E.fromEnum(E.toEnum(n))` must equal `n`
+1. `E.fromEnum(E.a)` is a natural number (countability)
+2. `E.toEnum(E.fromEnum(E.a)) === E.a` (bijection)
 
 #### `fromEnum` and `toEnum` functions
 
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 This project specifies interoperability of common algebraic
 structures:
 
+* [Enum](#enum)
 * [Setoid](#setoid)
 * [Ord](#ord)
 * [Semigroup](#semigroup)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,6 +92,37 @@ a `constructor` property which is a reference to the type representative.
...
 
+### Enum
+
+1. `E.fromEnum(E.a)` is a natural number (countability)
+2. `E.toEnum(E.fromEnum(E.a)) === E.a` (bijection)
+
+#### `fromEnum` and `toEnum` functions
+
+```hs
+fromEnum :: Enum a => a -> Natural
```
safareli: 0 is not Natural number

README.md
```diff
@@ -91,6 +92,37 @@ a `constructor` property which is a reference to the type representative.
...
+
+Given a value `a`, one can access its type representative via the
+`constructor` property:
+
+    a.constructor.fromEnum(b)
+    a.constructor.toEnum(n)
+
+1. `b` must be of the same Enum
+
+    1. If `b` is not the same Enum, an exception must be thrown.
```
safareli: Why exception instead of behavior being unspecified?

README.md
gabejohnson: That's up for debate http://mathworld.wolfram.com/NaturalNumber.html

README.md
```diff
@@ -91,6 +92,37 @@ a `constructor` property which is a reference to the type representative.
...
+### Enum
+
+1. `E.fromEnum(E.a)` is a natural number (countability)
+2. `E.toEnum(E.fromEnum(E.a)) === E.a` (bijection)
+
+#### `fromEnum` and `toEnum` functions
+
+```hs
+fromEnum :: Enum a => a -> Natural
+toEnum :: Enum a => Natural -> a
```
gabejohnson: Throw an exception.

README.md
gabejohnson: I think I copied this from the Haskell report. Should probably stay consistent with the other algebras

README.md
gabejohnson: Could leave it unspecified too though

ivenmarquardt:
I've read complaints about the `Enum` type class from Haskellers every now and then, because `Enum` implementations may be bounded, i.e. represent partial orders. `toEnum`/`fromEnum` would be unsafe in this context. Considering that the `Ordering` instance of your `Enum` type class would be incompatible with `Array.prototype.sort`, we would already need a mapping anyway. So why not define:
```
toEnum :: Natural -> Option a
fromEnum :: a -> Option Natural
```
so that that bounded enumerations are safe:
```
toEnum :: Natural -> Option (Bounded a)
fromEnum :: Bounded a -> Option Natural
```

gabejohnson:
@ivenmarquardt Fantasy Land has no definition of `Option` or `Maybe` types.

> ...the Ordering instance of your Enum type class would be incompatible with Array.prototype.sort...

Do you mean off-by-one? That's a simple conversion anyway.

I agree that `toEnum`/`fromEnum` are unsafe, but until it's decided that types can be defined in FL there's not much to be done about it.

PR #244, opened 2017-04-09
Complete the implementations for primitive JS types

i-am-tom:
Hello! I noticed the `internal/patch.js` file was looking a bit neglected, so I've added in all the sensible implementations I could think of for the primitive types. There are at least a couple we can't add because of limitations around type inference, but I left comments to point them out.

I'd also like to open the discussion as to whether `Object` can be considered a `Profunctor` type by the given implementation. I'm concerned that the keys in native JS objects aren't as flexible as one would hope, and you're obviously going to get funny behaviour with non-primitive types due to pointer lookups and other ugliness. I'm not optimistic, which is a shame :(

I've gone through every one and tested it in the console (and consequently fixed the `chain` implementation for `Array`, which was actually `map`). What do people think? :)

### DIFF
--- a/internal/func.js
+++ b/internal/func.js
@@ -1,13 +1,23 @@
 'use strict';
 
 const fl = require('../');
-const equality = (x, y) => typeof x[fl.equals] === 'function' ? x[fl.equals](y) : x === y;
+
+const equality = (x, y) =>
+  typeof x[fl.equals] === 'function'
+  ? x[fl.equals](y)
+  : x === y;
+
 const lte = (x, y) => {
-  if (typeof y[fl.lte] === 'function') return y[fl.lte](x);
+  if (typeof x[fl.lte] === 'function') {
+    return x[fl.lte](y);
+  }
 
   const typeX = typeof x;
   const typeY = typeof y;
-  return typeX === typeY && (typeX === 'string' || typeX === 'number') && x <= y;
+
+  return typeX === typeY
+    && ['string', 'number'].indexOf(typeX) !== -1
+    && x <= y;
 };
 
 module.exports = {
--- a/internal/patch.js
+++ b/internal/patch.js
@@ -3,31 +3,168 @@
 const fl = require('..');
 
 module.exports = () => {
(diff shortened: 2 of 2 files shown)


i-am-tom:
Apparently, the tests for `Ord` weren't actually testing order. Just fixing now!

Also not 100% on `chainRec` for these types. I'm guessing the `Id` one will suffice for `Array`:

```javascript
Array.prototype[fl.chainRec] = function(f, i) {
  const next = v => ({ done: false, value: v });
  const done = v => ({ done: true, value: v });

  let state = { done: false, value: i };

  while (state.done === false) {
    state = f(next, done, state.value).value;
  }

  return state.value;
}
```

davidchambers:
My hope is that once we merge fantasyland/fantasy-laws#1 we'll be able to remove the two files you've updated in this pull request (#174).

Thanks to sanctuary-type-classes it isn't necessary to mutate built-in prototypes in order to test conformance. You may like to submit a pull request to sanctuary-type-classes if it's missing some implementations, so your work will not be lost when we remove the assertion functions from this project.

i-am-tom:
Oh :( So I wasted my time?

davidchambers:
> So I wasted my time?

I imagine it was a useful learning experience, if nothing else. :)

I believe you misunderstood the purpose of __internal/patch.js__, though. It was not intended to provide implementations for all the built-in types. Rather, it was a hack to get some tests to pass. We wanted to show that `Identity a` has a semigroup if `a` has a semigroup, so we modified `String.prototype` so we could concatenate `Identity('abc')` and `Identity('def')` to get `Identity('abcdef')`.

i-am-tom:
Unfortunately, these are all things I've implemented many times, hence my excitement to find a reusable place to put them, hah! Ah well. C'est la vie.

davidchambers:
> hence my excitement to find a reusable place to put them

Is sanctuary-type-classes not that place?

i-am-tom:
Oh, I've no doubt! I'll PR anything (_if_ there's anything) that's missing from `sanctuary-type-classes` over there when I next get some free time :D

PR #248, opened 2017-05-16
Add Semigroupoid and Category algebras

gabejohnson:
Closes #208.

This PR adds the Semigroupoid and Category Algebras to the spec. They consist of two new methods: `compose` and `id`.

PR #249, opened 2017-05-17
Flip arguments in compose type signature

gabejohnson:
In PureScript
```hs
> append [1] [2] == [1] <> [2]
true

> (compose (_ + 1) (_ / 3)) 4 == ((_ + 1) <<< (_ / 3)) 4
true
```
I think I made a mistake with the signature for `compose`. It should be
```hs
Semigroupoid c => c j k ~> c i j -> c i k
```
not
```hs
Semigroupoid c => c i j ~> c j k -> c i k
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ A value which has an Ord must provide a `lte` method. The
 #### `compose` method
 
 ```hs
-compose :: Semigroupoid c => c i j ~> c j k -> c i k
+compose :: Semigroupoid c => c j k ~> c i j -> c i k
 ```
 
 A value which has a Semigroupoid must provide a `compose` method. The


davidchambers:
I'll let the other two comment on this change. My role in #248 was administrative in nature. ;)

safareli:
in FL

``` haskell
map :: f a ~> (a -> b) -> f b
ap :: f a ~> f (a -> b) -> f b
compose :: c i j ~> c j k -> c i k
```

in PS
``` haskell
map :: (a -> b) -> f a -> f b
apply :: f (a -> b) -> f a -> f b
compose :: a c d -> a b c -> a b d
```

```
f `map` val
map f val
val.map(f)

g <<< f
g `compose` f
compose g f
f.compose(g)
```

So in FL we have arguments flipped

gabejohnson:
@safareli do you agree that it is a bug that should be fixed?

safareli:
No what i'm saying is that if FL we all of functions have signature flipped so it's fine

gabejohnson:
I suppose I had `concat` in mind when I questioned the ordering.

```
a <> b
a `append` b
append a b
a.concat(b)
```
`Array.prototype.concat` matches the the ordering of PureScript for lists (an exception to the general rule). However, this isn't dictated by the Semigroup spec.

Were `Function.prototype.compose` to be added to ECMA-262 I suppose it would be more ergonomic using the current ordering.

Closing.

PR #252, opened 2017-06-09
Require that Profunctor implement the Contravariant spec

gabejohnson:
Values which implement Bifunctor or Profunctor must also implement Functor.

Profunctor is covariant in the second argument and Bifunctor in both.

Since Profunctor is contravariant in the first argument, shouldn't a value implementing it also be required to implement Contravariant?


gabejohnson:
I realize the graph looks funky now. I'll try to fix that.

davidchambers:
This sounds completely reasonable to me. I'll defer to those with better understandings. :)

:warning: This is a breaking change, so will warrant incrementing the major version number if merged.

gabejohnson:
@davidchambers I'm no longer sure that this is valid. I think (based on a comment by @puffnfresh on Gitter) that this reasoning works for Bifunctor and Profunctor with Functor because fixing the first argument results in a type that  _has_ a functor.
```
:k Bifunctor
Bifunctor :: (* -> * -> *) -> Constraint
```
```
 :k Functor
Functor :: (* -> *) -> Constraint
```
Now this isn't the case with Profunctor because it would be required to fix the second argument. Though honestly my type foo isn't sufficient to be certain.

rjmk:
It doesn't lead to inconsistency (though it breaks type inference), to allow type level lambdas like `instance Contravariant (Λt -> t -> a) where ...` (here the idea is to have the analogue to `(->) a`). It would quite a big change to the way Fantasy Land does things though

davidchambers:
What's the status of this pull request, @gabejohnson?

PR #257, opened 2017-07-12
Fix Apply "ap" method spec

1ven:


README.md
```diff
@@ -378,7 +378,7 @@ method takes one argument:
 
 1. `b` must be an Apply of a function
 
-    1. If `b` does not represent a function, the behaviour of `ap` is
+    1. If `b` does not represent an Apply of a function, the behaviour of `ap` is
        unspecified.
```
davidchambers: I don't know why we're using _represent_ here. This would be clearer to me:

_If `b` is not an Apply of a function, the behaviour of `ap` is unspecified._

Avaq: I think saying that `b` must be **an Apply of** a function is double, as we already stated that right before:

> 1. `b` must be **an Apply of** a function

I read *represent* to mean *the value represented by the Apply*, or *the inner value*. So in my opinion, the original sentence was alright.

Avaq: Maybe

*If `b` contains something other than a function, the behaviour of `ap` is unspecified.

davidchambers: If I read _represent_ as _wrap_, the current wording seems reasonable.

This implies that there's exactly one function "inside", which is not necessarily the case: `[]` and `[Math.sin, Math.cos]` are straightforward counterexamples.

davidchambers:
I'm closing this pull request as I don't think this is necessarily the right change. I'm sure this sentence can be clarified, but the exchange above did not produce a clear improvement.

Issue #262, opened 2017-07-20
functions without parameters


xgbuils:
Hi,

I've been reading [Type signature notation]() and the definition of `->` function type constructor.

But reading about laws specification I found something like:

- `empty :: Monoid m => () -> m`

However, something like `() -> a` does not satisfy the definition of type constructed by function type constructor:

Notice that say:
> When used in this way, the syntax is: (<input-types>) -> <output-type>, where <input-types> comprises **two or more comma–space** (,)-separated type representations.

If I'm not mistaken, the definition of function type constructor should be redefined.

davidchambers:
<https://sanctuary.js.org/#types> is accurate. We should copy its explanation.

proc0:
I don't think that's a function type constructor, if I understood correctly. `empty` is a function interface that monoids implement. In haskell it would be:
```haskell
class Monoid m where
  mempty :: m
```
However I noticed that fantasy-land has adapted function types for javascript, since function signatures in js are different than in haskell. So `() -> a` just means that you have a no-argument function that gives you a the 'empty element' of that monoid. 
```haskell
instance Monoid [a] where
   mempty = []
```

xgbuils:
I can do a PR extending the definition puting `zero or more comma–space...` instead of `two or more comma–space...`. Then, `(a) -> b` and `a -> b` represents the same type. Is right for you?

davidchambers:
That would be great, @xgbuils. :)

PR #263, opened 2017-07-22
Add 0 arguments Function type constructor definition

xgbuils:
related to #262


davidchambers:
I think it's clearer to state that the general form is `(<input-types>) -> <output-type>` but that the parens may be omitted for unary functions.

xgbuils:
I will do another PR with the suggested explanation.

Thanks!

PR #266, opened 2017-07-26
Bake in TypeScript && ES6 module compatibility

bcherny:
Before, there were no TypeScript typings, and to use FL in a strict ES6 module environment (eg. TypeScript) you had to do:

```ts
import * as fl from 'fantasy-land'
const {ap, chain, map, of} = fl
```

Now, you can do:

```ts
import {ap, chain, map, of} from 'fantasy-land'
```

~~Additionally, this PR:~~

- ~~Checks in package-lock.json (see [NPM docs](https://docs.npmjs.com/files/package-lock.json))~~
- ~~Adds a .npmignore for much faster installs~~
- ~~Adds .DS_Store to .gitignore~~

--------

- [x] Open question: Is there a need to maintain backwards-compatibility with the `window.FantasyLand` namespace, or is it fine to drop support for it?


davidchambers:
I appreciate the one-line saving, but there's no way this justifies adding two-and-a-half-thousand lines to this project. ;)

> Is there a need to maintain backwards-compatibility with the `window.FantasyLand` namespace, or is it fine to drop support for it?

Let's continue to support this.

bcherny:
@davidchambers 

> Let's continue to support this.

Done.

> there's no way this justifies adding two-and-a-half-thousand lines to this project. ;)

That's the package-lock.json. I can remove it, but it should be checked in for reproducible builds (see [NPM docs](https://docs.npmjs.com/files/package-lock.json)).

davidchambers:
> > there's no way this justifies adding two-and-a-half-thousand lines to this project. ;)
>
> That's the package-lock.json. I can remove it, but it should be checked in for reproducible builds (see [NPM docs](https://docs.npmjs.com/files/package-lock.json)).

Reproducible builds are very important for applications. I don't see value in including lock files for libraries, though. Please remove the lock file.

.gitignore
```diff
@@ -1 +1,3 @@
-node_modules
\ No newline at end of file
+.DS_Store
```
davidchambers: Only artefacts created by a project should be included in the project's __.gitignore__, in my view. Otherwise every project's __.gitignore__ will eventually contain hundreds of lines of noise relating to various operating systems and text editors. It's better for macOS users to add `.DS_Store` to their [global __.gitignore__][1]. :)


[1]: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

tslint.json
```diff
@@ -0,0 +1,45 @@
...
+      "call-signature": "nospace",
+      "index-signature": "nospace",
+      "parameter": "nospace",
+      "property-declaration": "nospace",
+      "variable-declaration": "nospace"
+    }],
+    "use-isnan": true,
+    "whitespace": [true, "check-branch", "check-decl", "check-module", "check-operator", "check-separator", "check-type"]
+  }
+}
```
davidchambers: Please remove this file.

bcherny:
> Reproducible builds are very important for applications. I don't see value in including lock files for libraries, though. Please remove the lock file.

The idea is to have reproducible builds of the library. Otherwise, the build might fail on your machine and on CI, but succeed on my machine (or some permutation of that). I removed it per your request, but be aware that you might run into issues at some point.

If you're worried about `npm install` size, be aware that the lockfile is automatically ignored when installing.

tslint.json
bcherny: Done. I was trying to keep it in line with the existing eslint step.

.gitignore
bcherny: Done, though in my experience these have never accumulated :)

.npmignore
```diff
@@ -0,0 +1,6 @@
+.travis.yml
+figures
+logo.png
+node_modules
+test.js
+*.md
```
davidchambers: This file is unnecessary due to the `"files"` field in __package.json__, I believe.

davidchambers:
Are supporting TypeScript and ES6 module syntax interrelated concerns? I ask because I believe the implications of these changes would be easier to grasp were there two pull requests rather than one.

.npmignore
bcherny: You're right - good catch!

bcherny:
They are separate, but TS needs ES6 exports. The only TypeScriptism here is the typings (eg. `: 'fantasy-land/equals'`). The rest is just ES modules. Before there was a single object with a bunch of keys being exported, now there are a bunch of strings being exported (no object). That's it.

davidchambers:
@bcherny, I've opened a competing pull request, #270. I'm interested in your thoughts on it.

bcherny:
@davidchambers Is there something wrong with this PR? It's really discouraging to have spent several hours on a PR, gone back and forth amending it, then having a contributor decide to make their own version instead :(

davidchambers:
> It's really discouraging to have spent several hours on a PR, gone back and forth amending it, then having a contributor decide to make their own version instead :(

I'm sorry that my actions had this effect, @bcherny.

I'm just one of several maintainers of this project, and I don't mean for #270 to supersede this one. It's an alternative approach worth considering.

Let me explain my motivations. I contribute to a significant number of open-source projects in my free time. To make this manageable I have spent time and energy setting up reusable infrastructure ([doctest][1], [sanctuary-style][2], [transcribe][3], [xyz][4]).

I believe that if one does not actively fight entropy a project will eventually be superseded by an alternative without the constraints of backwards compatibility, with a simpler interface, and with lightweight infrastructure.

Fantasy Land is first and foremost a specification. The decision was made to also provide a module which exports some strings. The code component of Fantasy Land is essentially 20 strings (setting aside `fantasy-land/laws` which will become a separate package).

Trivial code deserves lightweight tooling. This is why I balked at the initial size of this pull request. Pleasingly, this pull request is now much smaller, but it still introduces dependencies:

```console
$ rm -rf node_modules
$ npm install rollup@^0.45.2 rollup-plugin-typescript@^0.8.1 typescript@^2.4.2
$ find node_modules -name '*.js' -print0 | xargs -0 wc -l | tail -n 1
  652660 total
```

Half a million lines of code seems out of proportion to the difficulty of the problem we are solving.


[1]: https://github.com/davidchambers/doctest
[2]: https://github.com/sanctuary-js/sanctuary-style
[3]: https://github.com/plaid/transcribe
[4]: https://github.com/davidchambers/xyz

gabejohnson:
@bcherny I get where you're coming from, but if your goal was to get TS support in FL then it looks to me like your PR will be successful (even if it isn't merged).

I doubt David would have submitted #270 had you not both collaborated on _this_ PR :smile:

bcherny:
I'm not sure if I understand everything happening in that PR, but I have to say it's strange to avoid using TSC when generating TypeScript definitions. The toolchain in this PR (TSC + Rollup) is really standard (you could similarly use TSC + Browserify or TSC + Webpack). The file size is just for dev builds - I'm not sure I understand why you would opt for a hand-crafted build script over the industry standard build tools. Here are the pros/cons of using a custom build tool from what I can make out:

Pros:

1. Small file size when developing
2. Less NPM dependencies

Cons:

1. Does not scale to additional transpiled code in the future
2. Likely has bugs, which you will have to spend time fixing
3. Harder for others to contribute to

That said, the issue with your TypeScript code is you're typing all of the exports as `string`. Instead, you want string literal types (eg. `'fantasy-land/concat'`, see the code in my PR for more examples).

davidchambers:
> That said, the issue with your TypeScript code is you're typing all of the exports as `string`. Instead, you want string literal types […].

I'm intrigued by TypeScript's literal types, and I'll freely admit that I haven't come to terms with them yet. Could you provide an example of an invalid program which TypeScript would disallow as a result of knowing that the type of `FL.concat` is `'fantasy-land/concat'` rather than simply `string`?

andnp:
It would be nice to see this or #270 merged. I'm trying to use fantasy-land to make a TS project compliant with the spec, but would have to maintain my own types manually to do so.

Re: `'fantasy-land/concat'` vs `string`, I would be in favor of the more restrictive type for two reasons. 

First, I _always_ try to make sure my types are as specific as possible, and that they include the most amount of information possible. Because I can't always guess the way the types will be used by consumers, I want to maximize the information they contain since it is always easier to remove information than to add it.

Second, in my particular use-case `string` would be just as useless as not having types at all. I would like to define a class like:
```typescript
class Maybe<T> {
  [fl.map] = this.map.bind(this);
}
```
If `fl.map` was type `string`, then now an instance of my class could do:
```typescript
const m = new Maybe('thing');
m["I'm some arbitrary string"](v => v.toUpperCase())
```
or more modern versions of TS would simply disallow `[fl.map]` inside the class because "[it's not] a literal type or a 'unique symbol' type".

Anyways, if there is anything I can do to help move this issue along, let me know. I'm more than happy to help contribute to this project.

davidchambers:
Folks, I just submitted #294. Please have a look. :)

davidchambers:
This pull request was superseded by #294.

Thanks for prompting us to improve our TypeScript support, @bcherny. :)

Issue #269, opened 2017-08-10
Any interest in adding a `Group` spec?


evilsoft:
## Groups for all?
![image](https://user-images.githubusercontent.com/3665793/29182661-95b622da-7db4-11e7-89f6-5d15fdce7375.png)

While it is really strict and probably not grabbed for a lot, I was wondering if it would be beneficial to add a `Group` spec with the following bits and bobs:

### Group

A value that implements the Group specification must also implement
the [Monoid](#monoid) specification.

1. `g.concat(g.inverse())` is equivalent to `g.empty()` (right inverse)
2. `g.inverse().concat(g)` is equivalent to `g.empty()` (left inverse)

#### `inverse` method

```hs
inverse :: Group g => g ~> () -> g
```

evilsoft:
Actually. Maybe because it is on an instance, it should be more VERB-y and go with `invert` like Haskell.

evilsoft:
Well one person is good enough for me!!
Submitted [this PR](https://github.com/fantasyland/fantasy-land/pull/272) to address this issue.

PR #272, opened 2017-08-29
Add `Group` specification

evilsoft:
## A little `Group` therapy for the masses
image

This PR addresses [this issue](https://github.com/fantasyland/fantasy-land/issues/269) by adding a specification for `Group`. In addition to the new spec, I did the following to get the tests to work using "mocks" already in place:
* Moved the `internal/string_sum` to `internal/sum` and modified it to work with `Number` instead of `String`.
* Had to modify the `monoid` and `traverable.composition` specs to accommodate the newly `Number` dependent`Sum`.
* Fixed up the `semigroupoid` spec to run. The spec was not included in the specs, it is now.


README.md
```diff
@@ -306,6 +307,26 @@ Given a value `m`, one can access its type representative via the
 
 1. `empty` must return a value of the same Monoid
 
+### Group
+
+A value that implements the Group specification must also implement
+the [Monoid](#monoid) specification.
```
rjmk: [There's more than one way to get to group](https://en.wikipedia.org/wiki/Magma_(algebra)#/media/File:Magma_to_group2.svg), but we definitely shouldn't let that get in the way of this PR

(We could eventually have

```
class Magma m where
  concat :: m -> m -> m

class Magma m => Quasigroup m where
  latin :: (m, m) -> (m, m)

class Magma m => Unital m where
  empty :: () -> m

class Magma m => Semigroup m

class (Semigroup m, Unital m) => Monoid m
class (Quasigroup m, Unital m) => Loop m
class (Semigroup, Quasigroup, Unital m) => Group m
```

All (non-empty) associative quasigroups are groups, so there's no 8th member)

evilsoft: Oh snap. I see some possible opportunities here for another couple PRs!
Good call!

evilsoft:
@gabejohnson 🏫 me on how to get dat png updated.
Will get that in there today and take the WIP of of this.

evilsoft:
Also, I am not satisfied with the names `rightInverse` and `leftInverse` any suggestions for better/established names for those?

evilsoft:
Okay. Removed *WIP*, this is GTG and 🔴 👁 for review.

figures/dependencies.dot
```diff
@@ -42,7 +43,7 @@ digraph {
   Functor -> Profunctor;
   Functor -> Traversable;
   Plus -> Alternative;
-  Semigroup -> Monoid;
+  Semigroup -> Monoid -> Group;
```
joneshf: Please move this to a separate line, alphabetically please.

```dot
Functor -> Traversable;
Monoid -> Group;
Plus -> Alternative;
Semigroup -> Monoid;
```

evilsoft:
@joneshf addressed

figures/dependencies.dot
```diff
@@ -42,8 +42,9 @@ digraph {
   Functor -> Extend;
   Functor -> Profunctor;
   Functor -> Traversable;
+  Monoid -> Group;
```
joneshf: 👍 Thanks!

evilsoft:
Any other feedback or anything I should address that is keeping this from being merged in?

xaviervia:
What's the status of this one? It looks good :D

joneshf:
Agreed. We can follow up if something is missing/wrong.

gabejohnson:
A little late to the party. I think there's a small mistake

1. `g.concat(g.invert())` is equivalent to `g.empty()` (right inverse)
2. `g.invert().concat(g)` is equivalent to `g.empty()` (left inverse)

I think it should be either `g.constructor.empty()` or `G.empty()`.

See https://github.com/fantasyland/fantasy-land#plus

davidchambers:
Good catch, @gabejohnson. See #275.

PR #274, opened 2017-10-08
add Filterable algebra

davidchambers:
Closes #33

While working on this pull request I consulted both [Data.Witherable][1] and [Control.Compactable][2]. Neither definition is appropriate for Fantasy Land due to our decision not to rely on data types not provided by the language. We must use `a -> Boolean` rather than `a -> Maybe a`.

While reviewing this pull request please consider the following questions:

  - Should Filterable depend on any type classes other than <del>Monoid</del> <ins>Plus</ins>?
  - Can we state any other laws?

I'm requesting a review from you, @joneshf, as way back in sanctuary-js/sanctuary#10 you planted the seed that resulted in this pull request. :seedling:

/cc @Fresheyeball

[1]: https://hackage.haskell.org/package/witherable-0.1.3.4/docs/Data-Witherable.html
[2]: https://hackage.haskell.org/package/compactable-0.1.0.2/docs/Control-Compactable.html

PR #278, opened 2017-11-30
add Compactable algebra

### LINKED ISSUE #33: Filterable algebra?
`.map` and `.chain` are cool for writing higher order functions over things.

Another primitive I use a lot is `.filter` and I don't know how to implement in terms of `.map` / `.ap` / `.chain`.

Is there an algebra for it?

davidchambers:
Closes #33

This is an alternative to #274, based around `compact` rather than `filter`. `compact` refers to `Maybe a`, so this pull request defines what is meant by _a value of type `Maybe a`_. I opted for “object encoding” as suggested by @paldepind, as it seems more idiomatic in a JavaScript setting.


Fresheyeball:
Good job!

davidchambers:
> Find throwing me a mention for this?

Adding a link to [Control.Compactable][1] do you mean, @Fresheyeball? That sounds good to me.


[1]: https://hackage.haskell.org/package/compactable-0.1.0.2/docs/Control-Compactable.html

README.md
```diff
@@ -769,6 +804,41 @@ The `profunctor` method takes two arguments:
...
+`{isJust: false}` represents `Nothing`; `{isJust: true, value: 42}` represents
+`Just 42`.
+
+### `Nothing`
+
+`Nothing` is defined as:
+
+```js
+//    Nothing :: Maybe a
+const Nothing = {isJust: false};
```
rjmk: Will this be 'fantasyland/isJust'? My reading of the spec is that it will be (and I think this should be the case), but I wanted to clarify

gabejohnson: The API of `Maybe` as defined here isn't intended to be part of the specification. It could just as easily be defined as
```js
const Just = x => n => j => j (x);
const Nothing = n => j => n;
const isJust = m => m (false) (x => true);
const fromJust = m => m () (x => x);
```
The the definition would be:

A JavaScript value `x` is of type `Maybe a` if:
 - `isJust(x)` is `false`; or
 - `isJust(x)` is `true` and `fromJust(x)` is of type `a`.

Please correct me if I'm wrong on this point @davidchambers

rjmk: It could be defined that way, but it's not. If the definition is not in the spec, how can implementers actually write `compact`?

gabejohnson: Any conforming library which implements `compact` would also have to have an implementation of `Maybe`. There is no requirement that it have `Maybe.prototype.isJust`, just that `compact` can tell the difference between `Nothing` and `Just a`.

rjmk:
Got it. I think that could be clarified

README.md
paldepind: @gabejohnson I may be misunderstanding what you're saying, but it sounds to me like you're saying that `compact` can return anything it wants to represent `Nothing` and `Just`? That can't be right. What `compact` returns _must_ have an `isJust` property. If I write code that will work with Compactable implementations from several different libraries then there needs to be a consistent way to tell the return values apart from each other. `const ret = compact(a); ret.isJust === true` gives me that.

The properties in the return value do not need to be prefixed since they are already inside a Fantasy Land method that is prefixed. I.e. the sole purpose of the method is to implement Compactable so returning the proper values for implementing Compactable should not be an issue.

> Any conforming library which implements compact would also have to have an implementation of Maybe.

It may do that—but not necessarily. For instance, if this PR gets merged I intend to make `Stream` in the FRP library [Hareactive](https://github.com/funkia/hareactive) implement Compactable. But Hareactive is an FRP library so it naturally does not contain a Maybe implementation. Instead the `compact` implementation will simply do something like this:

```
return somebool ? { isJust: true, value: a } : { isJust: false };
```

Of cause libraries that already have a Maybe implementation may want to add the `isJust` property to their Maybe implementation and then they could seamlessly use it in their `compact` implementations. That is a positive side-effect of the object representation that can give implementors an added convenience. But that can, in general, not be relied upon when working with an arbitrary `compact` method from an arbitrary library.

davidchambers: In order to directly compare “function encoding” with “object encoding” I will open a third pull request.

gabejohnson: @paldepind 
> but it sounds to me like you're saying that compact can return anything it wants to represent Nothing and Just?

`compact` shouldn't return _anything_ to represent a `Maybe` unless that's the `Compactable` it's applied to.

```js
> compact([Just(1), Nothing, Just(2)])
[1, 2]

> compact(Just(Just(1)))
Just(1)

> compact(Just(Nothing))
Nothing
```

The latter two examples _require_ an implementation of `Maybe` which defines a `fantasy-land/compact` method.

My understanding is that the `Maybe` in this PR is defined solely for the purposes of defining the behavior of `compact` and not to act as specification for an ADT.

@davidchambers please correct me if I'm mistaken.

gabejohnson: Note that I'm not against specifying ADTs in FL, I just didn't think that's what this PR was doing.

davidchambers: > My understanding is that the `Maybe` in this PR is defined solely for the purposes of defining the behavior of `compact` and not to act as specification for an ADT.

I'd very much like us to discuss a concrete example such as this:

```javascript
//  Cons#fantasy-land/compact :: List (Maybe a) ~> () -> List a
Cons.prototype['fantasy-land/compact'] = function() {
  return this.head.isJust ?
    Cons(this.head.value, Z.compact(this.tail)) :
    Z.compact(this.tail);
};
```

Using Simon's “object encoding”, determining whether `this.head` is a Just simply requires accessing its `isJust` property. If `this.head` is in fact a Just we can access its inner value via the `value` property.

Gabe, could you show how `Cons#fantasy-land/compact` would be implemented using “function encoding”? The thing I don't understand is what the method would provide as the folding function to `isJust`. It could provide [`S.maybe`][1], for example, but this would not work for all possible values of type `Maybe a`. What am I missing?


[1]: https://sanctuary.js.org/#maybe

gabejohnson: I don't particularly like the "function encoding", and wouldn't advocate an implementation use it. In your `Cons` example, `compact` would of course use the Sanctuary implementation of `Maybe`.

If we want to define an API for ADTs in FL, I think there should be a separate issue for it. It would be _great_ for library interop and `Maybe` would be a good first candidate. I just think it should be in a separate PR.

The only reason I mention "function encoding" is because @puffnfresh  is resistant to the idea of Fantasy Land's specifying ADTs.

davidchambers: > I don't particularly like the "function encoding", and wouldn't advocate an implementation use it. In your `Cons` example, `compact` would of course use the Sanctuary implementation of `Maybe`.

Doesn't this mean that `List#fantasy-land/compact` would only work with *Sanctuary* Maybe values? This seems problematic as it introduces tight coupling between ADTs.

Perhaps #274 is not so bad after all. It gets us many of the benefits of Compactable without the problems associated with an ADT dependency.

paldepind: @gabejohnson 

> compact shouldn't return anything to represent a Maybe unless that's the Compactable it's applied to.

Oh. Yes, you are absolutely right. I had not looked at the type of `compact` closely enough 😅  I mistakenly thought that the `Maybe` was in the return type of `compact` and not in the arguments. I apologize for the confusion and thank you for clearing it up 👍 

> I don't particularly like the "function encoding", and wouldn't advocate an implementation use it.

I definitely agree.

@davidchambers 

> Perhaps #274 is not so bad after all. It gets us many of the benefits of Compactable without the problems associated with an ADT dependency.

I think you may be right. In fact, the difference between them seems very small. If a structure is `Filterable` and a `Functor` then `compact` can be derived as

```
function compact(a) { return a.filter(isJust); }
```

So it seems that there is little difference in power.

gabejohnson:
I've been thinking this over, and I think we should just use the sum type we have (`Boolean`). Introducing types to the spec should probably be done in a separate PR. We can always revisit `Compactable` if `Maybe` is added to the spec.

So my vote's for #274.

davidchambers:
> We can always revisit `Compactable` if `Maybe` is added to the spec.

Indeed. :)

PR #284, opened 2018-02-11
remove modules for verifying laws

### LINKED ISSUE #174: proposal: remove laws directory from repository

davidchambers:
Closes #174

Commit message:

> There is now a separate package, [fantasy-laws][1], which serves this role.

[1]: https://github.com/fantasyland/fantasy-laws


gabejohnson:
Why are you moving `Id` but removing `Compose` and `Sum`? I'm assuming it's because `Id` is included as an example demonstrating how one might go about implementing the specification while `Compose` and `Sum` are only provided for use in the tests.

If this is the case, I would move `equality` and `lte` into `id.js` and remove the dependency on `sanctuary-type-classes` (which incidentally isn't included in `package.json`).

davidchambers:
> I'm assuming it's because `Id` is included as an example demonstrating how one might go about implementing the specification while `Compose` and `Sum` are only provided for use in the tests.

That was my thinking, yes. We could go one step further by removing the Id implementation and linking to [sanctuary-identity][1] as soon as sanctuary-js/sanctuary-identity#3 is merged (this pull request could remain open until then).

> If this is the case, I would move `equality` and `lte` into `id.js` and remove the dependency on `sanctuary-type-classes` (which incidentally isn't included in `package.json`).

In order for `Id#fantasy-land/equals` to have value-based semantics one must do one of two things:

  - use `Z.equals` in the implementation; or
  - in addition to dispatching to `fantasy-land/equals`, provide suitable definitions for built-in types:
      - `Null#fantasy-land/equals`;
      - `Undefined#fantasy-land/equals`;
      - `Boolean#fantasy-land/equals`;
      - `Number#fantasy-land/equals`;
      - `Date#fantasy-land/equals`;
      - `RegExp#fantasy-land/equals`;
      - `String#fantasy-land/equals`;
      - `Array#fantasy-land/equals`;
      - `Arguments#fantasy-land/equals`;
      - `Error#fantasy-land/equals`;
      - `Object#fantasy-land/equals`; and
      - `Function#fantasy-land/equals`.

The file will better serve its pedagogical role if `Z.equals` is used in the implementation of `Id#fantasy-land/equals`, as this is the approach we would like ADT authors to adopt. Better yet would be to link to a complete working implementation (once it is available).


[1]: https://github.com/sanctuary-js/sanctuary-identity

Issue #177, opened 2016-09-27
Readme separate to specification


JAForbes:
I've received a lot of feedback from devs while working on getting Mithril, Lodash and Rx to support Fantasy Land regarding the heady language used in the specification.  It is intimidating and creates the impression that it won't be helpful in their work.

I personally think using precise language in a specification is exactly the right thing to do.  But I do think it would be beneficial to explain what the library is, what the benefits are, who is using it, how to get started implementing it, related projects etc on the front page.

I've received a lot of feedback from non functional programmers saying a blog post I wrote really helped them see the value, and they then went on to implementing in their own libraries, joining the gitter etc.  I think this is great!  I believe we can get really far if we build a nice smooth on-ramp.

So I'm proposing a separate `readme.md` and `spec.md`.  The current readme would become the spec, and the new readme would be a document extolling the virtues of fantasy land, etc etc.

The readme would be more colloquial (but still as precise as possible) and focus more on "how" / "why" than "what".  

How does that sound?

For anyone interested, here is a few recent examples, but I've got this sort of feedback many many times:

![image](https://cloud.githubusercontent.com/assets/2615540/18865630/977c8da8-84df-11e6-9f94-34be27844886.png)

[Source](https://news.ycombinator.com/item?id=12501350)

![image](https://cloud.githubusercontent.com/assets/2615540/18865873/ce08e4d8-84e0-11e6-9b06-ed20bf183cbf.png)

[Source](https://twitter.com/james_a_forbes/status/779985779023949825)

![image](https://cloud.githubusercontent.com/assets/2615540/18865689/e843f7e4-84df-11e6-8c39-7c4fad4589c2.png)

[Source](https://github.com/ReactiveX/RxJS/issues/34#issuecomment-249425793)

davidchambers:
Excellent idea, James. I agree that the readme should explain *why* a specification for interoperability of common algebraic structures in JavaScript is important. I appreciate the simplicity of your proposed solution.

rpominov:
Yea, it seems like currently the way this repo is organized is targeting people with Haskell background coming to JS who want to do Haskell stuff in JS. And I think we should make it more approachable for people coming from JS background and just wanting to do FP in JS.

I think this talk https://www.youtube.com/watch?v=oYk8CKH7OhE by author of Elm is very related here.

JAForbes:
@rpominov I nearly linked to that exact talk!  :D

bergus:
[...]

PR #286, opened 2018-03-17
Unified specification

rpominov:
This proposal unifies Fantasy Land and Static Land. The unified specification is based on _modules_ (simple objects with functions that don't use `this`) as opposed to putting methods directly to values. Although values still have a single property, reference to a module, which makes the unified specification equivalent in flexibility to Fantasy Land.

I see following advantages of this approach:

1. We need only one property that is subject to name clashes.
1. Sometimes it's impossible to add even a single property to values, but we still can implement a compliant module, which we'll have to pass explicitly everywhere though.
1. Also, we may implement multiple modules for a single type. Although again only one can be referenced from values, and another we'll have to pass explicitly.
1. Since every implementation will have modules, it allows for a whole different programming style where we choose to always pass modules explicitly even though they can be referenced from values. This may have some advantages, for example, types are more explicit in code.

Also, this approach should be easier to works with when we want to use static type systems like TypeScript or Flow. To be honest, I don't know much about this area. But other people did a lot of research. @gcanti works on https://github.com/gcanti/fp-ts and https://github.com/gcanti/flow-static-land . @alexandru works on https://github.com/funfix/funfix and argued that static functions are much easier to work with than method here https://github.com/rpominov/static-land/issues/45#issuecomment-363084044 . To my shame, I still haven't looked into these projects. But I hope @gcanti and @alexandru can participate in the discussion here.

Previous discussions about this unification approach can be found here: https://github.com/fantasyland/fantasy-land/issues/199 https://github.com/rpominov/static-land/issues/45

I also want to say a big thank you to @briancavalier who came up with the idea of unification!

This PR comes with a huge diff and comments may become hard to follow if we discuss every single change here. I think it'll be better if we only discuss here whether we want this or not in general, and for particular changes create issues or PRs in the fork https://github.com/rpominov/fantasy-land .


paldepind:
I'm a big fan of this :+1: Really great work @rpominov.

One additional benefit of this, that I haven't seen mentioned elsewhere, is that implementing the modules (without the method) is fully compatible with tree-shaking.

Here's a question: Since there will only ever be one canonical method can we name the method simply `canonical` or `fantasy-land`? To me `fantasy-land/canonical` feels a bit weird. The prefix with a slash seemed to make more sense when there were many prefixed methods.

rpominov:
Thanks for the support @paldepind ! 

I don't have a strong opinion about the property name. You can open issue or PR on the fork and we'll see what other people think.

briancavalier:
🎉 Thank you, @rpominov and everyone involved, for this amazing work!  I think this is a big win for the community ... and somewhat selfishly, as a maintainer of FL and SL compliant libs, I'm also very looking forward to supporting a single unified spec.

alexandru:
Hi all, hi @rpominov,

I'm very happy to see this.

In addition to [Funfix](https://funfix.org), I also started [Funland](https://github.com/funfix/funland) 🙂 This was extracted from Funfix and represents essentially static types for `static-land`, the explicit goal being 100% compatibility with it.

When working with a language like TypeScript, providing static-land compatibility first is very natural, because you can describe those type classes with actual types. And having to maintain `Fantasy-Land` compatibility in addition to `static-land` is a pain.

puffnfresh:
Is this method used anywhere anywhere else in JavaScript?

rpominov:
@puffnfresh You mean the property name that we're reserving?

puffnfresh:
@rpominov no, I mean defining functions via something like `fantasy-land/x`. Does this have any prior work elsewhere in JavaScript?

rpominov:
Ah. I'm not sure, probably not. At least not in a popular project that I know of. But then, there are not so many specifications in JavaScript ecosystem.

The closest thing that comes to mind is Iterable. Where value has a `Symbol.iterator` property that references a function that returns an object with the `next()` method. It's not quite the same, but the level of indirection is similar.

Still, I would agree that this approach is not very idiomatic for JavaScript. Usually, interfaces are defined by putting methods directly on `prototype` — exactly what Fantasy Land does now :)

alexandru:
Type classes aren't idiomatic JavaScript though. This is new territory.

paldepind:
I've created a PR https://github.com/rpominov/fantasy-land/pull/5 for simplifying the name of the canonical property to just `canonical`.

axefrog:
Added some thoughts to @paldepind's notes in https://github.com/rpominov/fantasy-land/pull/5

Raiondesu:
So, uhm, what's the status of this PR?

Seems like many things pivot around this and https://github.com/rpominov/fantasy-land/pull/5. 😅

For example, https://github.com/fantasyland/fantasy-land/issues/177, https://github.com/fantasyland/fantasy-land/pull/315, https://github.com/mostjs/core/issues/239, https://github.com/briancavalier/creed/issues/177, and https://github.com/ReactiveX/rxjs/issues/4674 with https://github.com/ReactiveX/rxjs/issues/34 (as well as many other issues in other different fp projects) seem unresolved due to this PR having no updates (as well as not having https://github.com/fantasyland/fantasy-land/issues/177).

davidchambers:
I should get back to #315. I believe that proposal will unify the two specifications, @Raiondesu.

PR #288, opened 2018-04-29
equivalent promises

Bamieh:
Two promises are equivalent when they yield equivalent values *for equivalent inputs*.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ have dependencies on other algebras which must be implemented.
     The definition should ensure that the two values can be safely swapped out in a program that respects abstractions. For example:
     - Two lists are equivalent if they are equivalent at all indices.
     - Two plain old JavaScript objects, interpreted as dictionaries, are equivalent when they are equivalent for all keys.
-    - Two promises are equivalent when they yield equivalent values.
+    - Two promises are equivalent when they yield equivalent values for equivalent inputs.
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
 ## Type signature notation


rpominov:
What is input for a promise? Promises just contain values, there is no input.

Bamieh:
@rpominov So bare with me here, what is a promise? a promise is a function that takes a resolver function as input and calls it passing `resolve` and `reject` to the `resolver` as params. hence when calling:

```
var a = new Promise(function resolver(resolve, reject) {
   resolve('value');
});
```

The input is the `resolver` function here. and for two promises to be equally the same, the input `resolver` functions must be equivalent.

The Promise return must be equivalent as well, which is something similar to

```
Promise {<resolved>: "value"}
```

And the promise yield is the result returned in the `then` payload, this also must be equivalent.

Looking at v8 code, here is how a promise is defined:

```
var $Promise = function Promise(resolver) {
  if (resolver === promiseRaw) return;
  if (!%_IsConstructCall()) throw MakeTypeError('not_a_promise', [this]);
  if (!IS_SPEC_FUNCTION(resolver))
...
///src: https://chromium.googlesource.com/v8/v8/+/3.29.45/src/promise.js?autodive=0%2F#38
```


hence comes my definition of a promise input.

rpominov:
I think we're confusing the Promise constructor and promise instances. Yes, the *promise constructor* is a function that takes a resolver and returns a *promise instance*. And specification talks about promise instances.

Bamieh:
@rpominov yes indeed, as the only way to interact with a promise instance is to use call a new promise.  In this case, a promise instance is a plain object with differing `[[PromiseResult]]` internal slot values when they resolve.

I will close, thanks for your time.

Bamieh:
@CrossEye thanks, i'll have a look at it

PR #293, opened 2018-06-16
integrate sanctuary-scripts

davidchambers:
[sanctuary-scripts][1] brings together several packages upon which this project already depends: [eslint][2], [sanctuary-style][3], and [xyz][4]. It also performs a number of checks to promote good project hygiene (one of which prompted the inclusion of __CONTRIBUTING.md__ in this pull request).

[1]: https://github.com/sanctuary-js/sanctuary-scripts
[2]: https://eslint.org/
[3]: https://github.com/sanctuary-js/sanctuary-style
[4]: https://github.com/davidchambers/xyz

PR #270, opened 2017-08-11
derive index.js and index.d.ts from canonical list of method names

davidchambers:
Like #266 this pull request makes it possible to write the following TypeScript code:

```typescript
import {ap, chain, map, of} from 'fantasy-land';
```

It achieves this in a less invasive manner:

  - it adds a text file, __names__, which simply lists the names of the exports;
  - it adds a template file, __src/index.js__;
  - it derives __index.js__ from __names__ and __src/index.js__; and
  - it derives __index.d.ts__ from __names__.

PR #294, opened 2018-06-17
derive index.js and index.d.ts from canonical list of method names

davidchambers:
This pull request supersedes #270, which included a few unrelated changes which have since been rendered unnecessary (#293).

Like #270, this pull request avoids npm dependencies by using standard Unix programs to generate __index.d.ts__. The `awk` script is just a single line:

```bash
awk '{ print "export const " $0 ": \047fantasy-land/" $0 "\047;" }' names >index.d.ts
```

Unlike #270, this pull request uses [string literal types][1] in __index.d.ts__ rather than the far less specific [`string`][2] type. I'm not a TypeScript user so I would greatly appreciate a review by @bcherny, @andnp, @sunderhus, or anyone else in the community with TypeScript knowledge. :)

This pull request also updates __CONTRIBUTING.md__ with instructions for updating the various files, and updates the behaviour of `npm run lint` such that ignoring these instructions will result in an error.

[1]: https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types
[2]: https://www.typescriptlang.org/docs/handbook/basic-types.html#string


paldepind:
Looks good to me. The types are correct and given how simple they are generating them in the proposed manner seems reasonable.

bcherny:
@davidchambers Sounds good. I’m not sure I’d make the same decision, but thanks for taking the time to clarify! TS looks good to me. One note- it looks like the generated JS use ES5-style exports, but the generated TS uses ES6-style exports. If I’m reading that right, I’d suggest sticking to one or the other.

davidchambers:
> One note- it looks like the generated JS use ES5-style exports, but the generated TS uses ES6-style exports. If I’m reading that right, I’d suggest sticking to one or the other.

Restricting the JavaScript file to ES5 syntax enables it to be used as is in environments (such as Node) which don't support `import`/`export`. Given that TypeScript files must necessarily be processed before being evaluated, is there a reason not to use `export` and `const` in that context?

bcherny:
@davidchambers It's just another small deviation from standard TS practice (declare in TS + ES6, compile to JS + ES5 with tsc). Looks good to me!

davidchambers:
> I'm no shell scripting expert, but, is there any reason for not making the scripts `#!/bin/sh` POSIX compatible? As far as I can tell they don't use any bashisms.

I like to understand the code I write, but with shell scripts I'm not averse to cargo cult programming. ;)

This pull request has been open for a year and a half. Let's merge it. I would not oppose a pull request to change the shebangs, but I'm confident `#!/usr/bin/env bash` will not cause us any problems.

Issue #211, opened 2016-12-15
Promise equlity


paldepind:
The readme says:

> Two promises are equivalent when they yield equivalent values.

Is this really an "appropriate definition of equivalence for the given value"? Because then I don't understand what the criteria for equivalence is?

It seems to me that a promise that resolves in five minutes with the value "Tada" is not equal to a promise that resolves tomorrow afternoon with the value "Tada". I'd say that two promises are only equal if they resolve to equivalent values at the exact same times.

SimonRichardson:
I do get the confusion, but `new Promise() != new Promise()` so the only way to really know if the promises are truly the same are to see the outcomes (apart for testing the instances are the same). I personally think we could improve that sentence and mention something along the lines (draft, don't hate me), `Two promises are equivalent if they are referentially transparent for the outcome of the promises.` or more elegantly put.

bergus:
The problem is that resolution timing is inherently a side-effect. In any pure monad, we'd expect
```js
const ma = f();
const mb = g().chain(() => ma);
```
and
```js
const mb = g().chain(() => f());
```
to have the same result `mb` (though not necessarily the same execution efficiency). This does not work with promises if we consider their impure timing to be relevant, so we choose to ignore it in the equivalence relation.

SimonRichardson:
> Two promises are equivalent if they are referentially transparent.

Maybe it should say that then, because covers all edge cases. If you don't think it does then you need to read more.

safareli:
What if we remove that line? 

If we say `Two promises are equivalent if they are referentially transparent.` we can remove `promise` and  just state: `Two values are equivalent if they are referentially transparent`

paldepind:
@SimonRichardson

> I do get the confusion, but `new Promise() != new Promise()` so the only way to really know if the promises are truly the same are to see the outcomes (apart for testing the instances are the same).

Well, my point is that just seeing the outcomes is really not enough to know if they are truly the same. To do that you'd have to know both the outcome and the time of resolution.

@bergus 

> This does not work with promises if we consider their impure timing to be relevant, so we choose to ignore it in the equivalence relation.

Then I think I'd be better to just say that Promises aren't monads because they don't satisfy the laws instead of adopting an IMO too loose sense of equality.

@safareli 

I think that is a great idea. It's completely general. However, the current text does already talk about referential transparency even though it doesn't use the term directly:

> The definition should ensure that the two values can be safely swapped out in a program that respects abstractions.

[...]

PR #295, opened 2018-06-17
readme: replace problematic definition of promise equivalence

davidchambers:
I've listed @safareli as co-author since he suggested this change in <https://github.com/fantasyland/fantasy-land/issues/211#issuecomment-267304326> back in 2016!

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ have dependencies on other algebras which must be implemented.
     The definition should ensure that the two values can be safely swapped out in a program that respects abstractions. For example:
     - Two lists are equivalent if they are equivalent at all indices.
     - Two plain old JavaScript objects, interpreted as dictionaries, are equivalent when they are equivalent for all keys.
-    - Two promises are equivalent when they yield equivalent values.
+    - Two values are equivalent if they are [referentially transparent](https://en.wikipedia.org/wiki/Referential_transparency).
     - Two functions are equivalent if they yield equivalent outputs for equivalent inputs.
 
 ## Type signature notation


paldepind:
I'm not sure how helpful the Wikipedia link is. The text in the readme say "Two values are equivalent if they are referentially transparent" but the definition in the Wikipedia article defines referential transparency as a property that an expression can have.

davidchambers:
Simon, how do you suggest we proceed?

Avaq:
```
2. "equivalent" is an appropriate definition of equivalence for the given value.
                                                                          ^^^^^
                                                                            1a

    The definition should ensure that the two values can be safely swapped out in a program that respects abstractions. For example:
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                                                      2a

    - Two lists are equivalent if they are equivalent at all indices.
    - Two plain old JavaScript objects, interpreted as dictionaries, are equivalent when they are equivalent for all keys.
    - Two values are equivalent if they are [referentially transparent](https://en.wikipedia.org/wiki/Referential_transparency).
          ^^^^^^                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
            1b                                          2b

```

I feel like we are using a circular definition of equivalence:

1. At `a` we talk about "value", and then proceed to give a list of examples of values. Yet in the examples we include a definition (at `b`) of arbitrary "value".
2. To my understanding, the sentence at `a` describes referential transparency. Yet in one of the examples of this (at `b`), we refer back to referential transparency.

paldepind:
I am not familiar with referential transparency as a term that defines equality between two values? The definition I'm familiar with is one similar to the one in the Wikipedia link which defines referential transparency as a property of purely functional programs. Can anyone point towards a definition of referential transparency that fits with how we're using the term here?

Here is an alternative that is precise and which doesn't use any jargon:

> "equivalent" means that two values can be safely swapped out in a program. Two values `a` and `b` are equivalent if for any pure predicate function `p` it is the case that `p(a)` evaluates to `true` if and only if `p(b)` evaluates to `true`. For example:
> * Two arrays are equivalent if they are equivalent at all indices.
> * Two plain old JavaScript objects are equivalent when they are equivalent for all keys.
> * Two promises are equivalent when they yield equivalent values at the same point in time.
> * Two functions are equivalent if return equivalent values for equivalent arguments.

The definition of equality I've formulated above is just [Liebniz equality](https://en.wikipedia.org/wiki/Equality_(mathematics)#Logical_formulations) which is both a very precise notion of equality and, in my opinion, one that is very intuitive and easy to understand.

davidchambers:
This pull request has proven controversial, and I'm not currently inclined to spend time and energy resolving the controversy. Feel free to submit a pull request if you believe you can satisfy all parties. :)

PR #304, opened 2018-11-25
Group invert behavior

ghost:
Missing behavior of `invert` method in `Group` added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ A value which has a Group must provide an `invert` method. The
 
     g.invert()
 
-1. `invert` must return a value of the same Group.
+1. `invert` must return a value of the same Group. If retrun value is not from the same Group, behaviour of invert is unspecified.
 
 ### Filterable
 


README.md
```diff
@@ -327,7 +327,7 @@ A value which has a Group must provide an `invert` method. The
 
     g.invert()
 
-1. `invert` must return a value of the same Group.
+1. `invert` must return a value of the same Group. If retrun value is not from the same Group, behaviour of invert is unspecified.
```
davidchambers:
```suggestion
1. `invert` must return a value of the same Group. If return value is not from the same Group, behaviour of `invert` is unspecified.
```

Avaq:
I don't think this change is valid. We don't consider the possibility for `invert` to return anything other than a value of the same Group. That possibility is only considered for *user supplied functions*. In the case of `invert`, if the return value is not from the same group, the implementation is simply not FL compliant.

evilsoft:
@remisa-yousefvand I wrestled with that as well, and added and deleted the same phrasing over and over again. And finally came to the same 🌽-clusion that @Avaq stated in their comment.

remisa-yousefvand:
@evilsoft I understand you and @Avaq point of view as a formal documentation where redundancy is prohibited. But if someone jumps in the middle of document to read about a specific item, redundancy may help a lot. That's the way ordinary ppl like me read documentations :book:

Avaq:
It's not just the redundancy I'm worried about. Making this change will cause it to seem as though the specification somehow tolerates `invert` implementations that deviate from the spec in terms of their output type.

`invert` must return a value of the same Group. Since it's up to the library author (as opposed to the library user), the specification considers only the valid implementation. An invalid implementation would deviate from the spec.

----

Note also:

> It is recommended to throw an exception on unspecified behaviour. -- [source](https://github.com/fantasyland/fantasy-land#notes)

When we're talking about the output type of an operation, that doesn't make sense - you either return a non-conformant value, or throw an exception, but not both.

I mention this in an attempt to make clearer the intent of the "... behaviour is unspecified" clauses.

If we examine the [`map` spec](https://github.com/fantasyland/fantasy-land#map-method), for example, we'll see in section **1.1** that "behaviour is unspecified" when `f` is not a function. That's because `f` is an argument *supplied by the user*, and since the user supplies it, the spec cannot strictly determine its type. In section **2**, however, the behaviour is assumed to always conform to the spec. That's because `map` is not supplied by the user, but rather a static method.

Avaq:
@remisa-yousefvand I agree that the spec is not very clear the way it's written. I think your idea of making the parts around "unspecified behaviour" in the spec more clear and consistent is good. It's just that I don't think this particular change is a step in the right direction.

It might be worth to consider making this kind of change instead:

```diff
 A value which has a Functor must provide a `map` method. The `map`
 method takes one argument:
 
     u.map(f)
 
 1. `f` must be a function,
 
-    1. If `f` is not a function, the behaviour of `map` is
-       unspecified.
     2. `f` can return any value.
     3. No parts of `f`'s return value should be checked.
+
+2. If a non-conformant value is supplied for `f`, the
+   behaviour of `map` is unspecified.
 
 2. `map` must return a value of the same Functor
```

That way, the spec is more clear about these values being "supplied", and specs for methods like [`ap`](https://github.com/fantasyland/fantasy-land/blob/master/README.md#ap-method) are also less inconsistent; I'm referring to the missing *"if `b` is not the same Apply as `a`, behaviour of `ap` is unspecified"* in section **1.2**.

davidchambers:
> It's not just the redundancy I'm worried about. Making this change will cause it to seem as though the specification somehow tolerates `invert` implementations that deviate from the spec in terms of their output type.

I find this argument compelling.

PR #311, opened 2019-02-18
Remove no part checked. Fixes #303.

### LINKED ISSUE #303: Is really necessary specify "No parts of <...> return value should be checked"?
Hi!

After reading #210, I still don't understand why these kind of law are needed. The topic starts talking about "No parts of `b` return value should be checked. But, finally, it concludes that the implementations that perform this kind of checks are breaking other laws. 

For example:
``` javascript
class Maybe {
    static of(val) {
      return new Just(val);
    }
}

class Just extends Maybe {
    constructor(val) {
        super()
        this.val = val;
    }
    map(f) {
        const r = f(this.val)
        return r === null ? new Nothing() : Maybe.of(r);
    }
}

class Nothing extends Maybe {
    map(f) {
        return this;
    }
}
```
This implementation breaks `No parts of f's return value should be checked`. However, it is also breaking the composition law:

``` javascript
const f = x => x === 2 ? null : x;
const g = x => x === null ? 2 : x;

Maybe.of(2).map(f).map(g) // Nothing {}
Maybe.of(2).map(x => g(f(x))) // Just {val: 2}
```

Then, I wonder if `No parts of f's return value should be checked` is a really necessary law. For me it's a weird law because is talking about the implementation. For example these two implementations of Maybe:

```
class Just extends Maybe {
    ...
    map(f) {
        return Maybe.of(f(this.val));
    }
}
```
and
```
class Just extends Maybe {
    ...
    map(f) {
        const r = f(this.val)
        return r === null ? Maybe.of(r) : Maybe.of(r);
    }
}
```
behave the same and are identity and composition law compliant. However, the second breaks the `No parts of f's return value should be checked` and is not functor lawful. It doesn't make sense for me. 

Cheers!

CrossEye:
This avoids the "No parts of <value> should be checked" from the individual rules and adds a `forAll` intro section to explain what that means.

It is done using `forall` annotations on signatures.  Could someone with deeper knowledge of H-M types double-check that I added it correctly?

### DIFF
--- a/README.md
+++ b/README.md
@@ -810,13 +810,13 @@ to implement certain methods then derive the remaining methods. Derivations:
   - [`map`][] may be derived from [`ap`][] and [`of`][]:
 
     ```js
-    function(f) { return this.ap(this.of(f)); }
+    function(f) { return this.ap(this.constructor.of(f)); }
     ```
 
   - [`map`][] may be derived from [`chain`][] and [`of`][]:
 
     ```js
-    function(f) { return this.chain(a => this.of(f(a))); }
+    function(f) { return this.chain(a => this.constructor.of(f(a))); }
     ```
 
   - [`map`][] may be derived from [`bimap`][]:
--- a/README.md
+++ b/README.md
@@ -159,6 +159,14 @@ Identity type, for example, could provide `Id` as its type representative:
 If a type provides a type representative, each member of the type must have
 a `constructor` property which is a reference to the type representative.
 
+
+## `forAll`
+
+[Parametric polymorphism][pp] is essential to these algebras.  When the rules for any type are specified as applied to all values (`forall a b`) of the given generic types (here `a` and `b`), the rule must hold for any Javascript values unless otherwise specified.  This implies, for instance, that code that behaves differently for the `null` or `undefined` value than for other values is unlikely to comply with the rule.
+
+  [pp]: https://en.wikipedia.org/wiki/Parametricity
+
+
 ## Algebras
 
 ### Setoid
@@ -363,7 +371,7 @@ method takes one argument:
 #### `map` method
 
 ```hs
-map :: Functor f => f a ~> (a -> b) -> f b
+map :: forall a b . Functor f => f a ~> (a -> b) -> f b
 ```
 
 A value which has a Functor must provide a `map` method. The `map`
@@ -376,7 +384,6 @@ method takes one argument:
     1. If `f` is not a function, the behaviour of `map` is
        unspecified.
     2. `f` can return any value.
-    3. No parts of `f`'s return value should be checked.
 
 2. `map` must return a value of the same Functor
 
@@ -389,7 +396,7 @@ method takes one argument:
 #### `contramap` method
 
 ```hs
-contramap :: Contravariant f => f a ~> (b -> a) -> f b
+contramap :: forall a b . Contravariant f => f a ~> (b -> a) -> f b
 ```
 
 A value which has a Contravariant must provide a `contramap` method. The
@@ -402,7 +409,6 @@ A value which has a Contravariant must provide a `contramap` method. The
     1. If `f` is not a function, the behaviour of `contramap` is
        unspecified.
     2. `f` can return any value.
-    3. No parts of `f`'s return value should be checked.
 
 2. `contramap` must return a value of the same Contravariant
 
@@ -416,7 +422,7 @@ implement the [Functor](#functor) specification.
 #### `ap` method
 
 ```hs
-ap :: Apply f => f a ~> f (a -> b) -> f b
+ap :: forall a b . Apply f => f a ~> f (a -> b) -> f b
 ```
 
 A value which has an Apply must provide an `ap` method. The `ap`
@@ -435,8 +441,6 @@ method takes one argument:
 3. `ap` must apply the function in Apply `b` to the value in
    Apply `a`
 
-   1. No parts of return value of that function should be checked.
-
 4. The `Apply` returned by `ap` must be the same as `a` and `b`
 
 ### Applicative
@@ -451,7 +455,7 @@ implement the [Apply](#apply) specification.
 #### `of` method
 
 ```hs
-of :: Applicative f => a -> f a
+of :: forall a . Applicative f => a -> f a
 ```
 
 A value which has an Applicative must provide an `of` function on its
@@ -467,8 +471,6 @@ Given a value `f`, one can access its type representative via the
 
 1. `of` must provide a value of the same Applicative
 
-    1. No parts of `a` should be checked
-
 ### Alt
 
 A value that implements the Alt specification must also implement
@@ -480,7 +482,7 @@ the [Functor](#functor) specification.
 #### `alt` method
 
 ```hs
-alt :: Alt f => f a ~> f a -> f a
+alt :: forall a . Alt f => f a ~> f a -> f a
 ```
 
 A value which has a Alt must provide a `alt` method. The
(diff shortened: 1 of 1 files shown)


davidchambers:
This is nice!

README.md
```diff
@@ -159,6 +159,14 @@ Identity type, for example, could provide `Id` as its type representative:
 If a type provides a type representative, each member of the type must have
 a `constructor` property which is a reference to the type representative.
 
+
+## `forAll`
+
+[Parametric polymorphism][pp] is essential to these algebras.  When the rules for any type are specified as applied to all values (`forall a b`) of the given generic types (here `a` and `b`), the rule must hold for any Javascript values unless otherwise specified.  This implies, for instance, that code that behaves differently for the `null` or `undefined` value than for other values is unlikely to comply with the rule.
```
davidchambers: Could you wrap this paragraph at 79 characters?

README.md
```diff
@@ -363,7 +371,7 @@ method takes one argument:
 #### `map` method
 
 ```hs
-map :: Functor f => f a ~> (a -> b) -> f b
+map :: forall a b . Functor f => f a ~> (a -> b) -> f b
```
davidchambers: PureScript appears not to include a space before the `.`. I suggest we follow suit.

README.md
davidchambers: Please change “Javascript” to “JavaScript” for consistency with the rest of the document.

README.md
```diff
@@ -159,6 +159,14 @@ Identity type, for example, could provide `Id` as its type representative:
 If a type provides a type representative, each member of the type must have
 a `constructor` property which is a reference to the type representative.
 
+
+## `forAll`
```
davidchambers: Might “Parametricity” be a better heading?

README.md
CrossEye: Done and done.

README.md
CrossEye: Fixed.

README.md
CrossEye: I don't think so.  If someone doesn't understand this signature because of the `forall` clause:

```js
map :: forall a b. Functor f => f a ~> (a -> b) -> f b
```

and tries to figure out what it is, a header that matches the confusing keyword is better than a term that -- given this confusion -- is likely unknown to them also.

README.md
safareli: (Space is allowed in PS before the dot btw)

README.md
```diff
@@ -159,6 +159,19 @@ Identity type, for example, could provide `Id` as its type representative:
...
+
+[Parametric polymorphism][pp] is essential to these algebras.  When the rules
+for any type are specified as applied to all values (`forall a b`) of the 
+given generic types (here `a` and `b`), the rule must hold for any JavaScript
```
safareli: I'm wondering how can we make restrictions polymorphism implies more clear here.
What you think on something like this:(bold is for highlighting the addition):

> Parametric polymorphism is essential to these algebras. When the rules
for any type are specified as applied to all values (`forall a b`) of the
given generic types (here `a` and `b`), the rule must hold for any **possible**
JavaScript value ~~unless otherwise specified~~. 
**Since a parametrically polymorphic value does not "know" anything about the
type variables, it must behave the same regardless of its type. This is a somewhat
limiting but extremely useful property known as [parametricity](https://en.wikipedia.org/wiki/Parametricity).
This implies, that a function should behave in same way for all values,
i.e. inspection of value is not allowed.** for instance, that code that
behaves differently for the `null` or `undefined` value than for other
values ~~is unlikely to comply~~ complying with the rule.


we can steal some ideas from https://wiki.haskell.org/Polymorphism or/and link to it too (has better explanation then wikipedia)

README.md
```diff
@@ -552,11 +559,7 @@ method takes two arguments:
     1. if `f` is not a function, the behaviour of `reduce` is unspecified.
     2. The first argument to `f` must be the same type as `x`.
     3. `f` must return a value of the same type as `x`.
-    4. No parts of `f`'s return value should be checked.
-
-1. `x` is the initial accumulator value for the reduction
```
safareli: why remove this line?

README.md
davidchambers: Good point. Shall we use `forall` rather than `forAll`, though?

joneshf:
Since we were writing Haskell-esque type signatures, we were also inheriting Haskell's implicit `forall` for all type variables. I.e. it was already implied that things were parametrically polymorphic. The bit about no values being checked was still added in even though the signature explained that part. And we still get implementations that are not law abiding. The fact that we've moved to explicit `forall` doesn't seem like enough of a justification to me to also remove the bit about no values being checked. Adding an explicit `forall` is good because it is acting like a springboard to explain parametric polymorphism!

I'm not going to veto this PR, I think people making changes is good and I support it. But, I don't think this solves the underlying problem of people writing unlawful implementations.

Also, if we're now writing PureScript-esque type signatures (or Haskell-esque with GHC's `ScopedTypeVariables` extension), then we need to be consistent. All of the type variables should be introduced in the `forall`. E.g. `map` should be:
```PureScript
map :: forall a b f. Functor f => f a ~> (a -> b) -> f b
```

not

```PureScript
map :: forall a b. Functor f => f a ~> (a -> b) -> f b
```

Being inconsistent will lead to further explanations needed about how GHC's `ScopedTypeVariables` extension actually works and that's far off the beaten path of this spec.

README.md
CrossEye: I have no problem putting it back.  I just didn't see it adding anything.  Is there something that makes it clear that an "empty" foldable will return this value?  Do you think that this text captures that?

CrossEye:
@joneshf:

The goal was not to add `forall`.  It was explicitly to remove "no values should be checked".  (See #303 for details.)  Adding `forall` seemed to be a compromise for those cases.  I have no problem removing `forall` everywhere and just depending upon an expanded section on parametricity.  I actually think that would be better.

README.md
CrossEye: Absolutely.  I'll fix that tonight.

README.md
CrossEye: I would like to expand this section.  These are good ideas.  The "unless otherwise specified" has to do with things like `f must be a binary function` or `b must be same Apply as a`.  In such cases, not every value is allowable.

"is unlikely to comply" was about examples such as  `return r === null ? Maybe.of(r) : Maybe.of(r)`.  The only way these examples could work is if their differences are not observable, so perhaps we can skip it.  But it was part of the motivation for this change.

README.md
CrossEye: Yes, allowed in Haskell as well, at least in some example code, but it's much more rare, it seems.

CrossEye:
@davidchambers: Rather than adding `forall` to the other signatures, let's look at removing all of them, and simply depending upon a parametricity section.  I'll try to create a separate PR tonight or tomorrow.

davidchambers:
@CrossEye, if you don't intend to update this pull request, shall we close it?

CrossEye:
I intend to create a replacement, but life has gotten in the way.  Hopefully soon.

puffnfresh:
Add `forall` to everything

xgbuils:
After @joneshf said that:

> Since we were writing Haskell-esque type signatures, we were also inheriting Haskell's implicit forall for all type variables.

What 's the reason to add forall to everything? In my opinion, it generates noise in the law definitions. A section which explains that all laws are parametrically polymorphic is enough for me.

I think we have to get an agreement in #303 before doing the PR. Because it seems that no part to be checked is too much opened to interpretations:

1. Does it mean parametric polymorphism?
2. Does it mean inspection?
3. Does it mean inspection with observable effect?

PR #316, opened 2019-10-08
Added Zion

oncicaradupopovici:
Zion is a functional programming library inspired by Haskell and Fantasy Land.

It provides some Haskell-like algebraic data-types and polymorphic functions against Fantasy Land algebraic structures.


davidchambers:
Thanks for the pull request. :)

PR #321, opened 2020-05-03
add es module file and declaration

TheLudd:
This adds a main file for es modules. Is is generated in the same way as
the ts types from the list of names.


TheLudd:
Done. Hope I found all the places to update.

Den mån 4 maj 2020 kl 01:57 skrev David Chambers <notifications@github.com>:

PR #328, opened 2021-07-31
Improve portability of the UMD build

chocolateboy:
This PR:

- replaces `self` with `globalThis`
- makes detection of CommonJS environments more robust

[`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) is the official way to refer to the global scope and is supported by almost all environments:

| engine          | self | globalThis |
|-----------------|------|------------|
| Node.js         | ✖    | ✔          |
| QuickJS         | ✖    | ✔          |
| SpiderMonkey    | ✖    | ✔          |
| Modern browsers | ✔    | ✔          |
| Deno            | ✔    | ✔          |
| IE11            | ✔    | ✖          |

The notable exception is IE11, which will need to use a [polyfill](https://github.com/search?q=globalThis+polyfill&s=stars). If IE11 support was previously promised or assumed, then this would be a breaking change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,10 +28,14 @@
   };
 
   /* istanbul ignore else */
-  if (typeof module === 'object' && typeof module.exports === 'object') {
+  if (
+    typeof module === 'object' &&
+    module != null &&
+    typeof module.exports === 'object'
+  ) {
     module.exports = mapping;
   } else {
-    self.FantasyLand = mapping;
+    globalThis.FantasyLand = mapping;
   }
 
 } ());
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,6 +10,10 @@
         "no-undef": ["off"],
         "no-unused-vars": ["off"]
       }
+    },
+    {
+      "files": "index.js",
+      "globals": {"globalThis": false}
     }
   ]
 }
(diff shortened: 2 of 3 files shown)


davidchambers:
Thanks for the pull request, @chocolateboy! I pushed a few tweaks to your branch. Are you happy with them?

chocolateboy:
Yes, LGTM, thanks!

index.js
```diff
@@ -28,10 +28,14 @@
...
-  if (typeof module === 'object' && typeof module.exports === 'object') {
+  if (
+    typeof module === 'object' &&
+    module != null &&
+    typeof module.exports === 'object'
+  ) {
     module.exports = mapping;
   } else {
-    self.FantasyLand = mapping;
+    globalThis.FantasyLand = mapping;
```
diasbruno, line 38: Wouldn't be safer to use something like `(globalThis || self)`? Just in case...

davidchambers: If `globalThis` does not exist in the environment, `globalThis || self` throws a ReferenceError. We would need to use `typeof globalThis` to first check for the presence of `globalThis`. Is this worthwhile? I'm not sure.

diasbruno: Yeah, you are right. I thought short-circuiting would not throw.

diasbruno: It'd be worth if it needs to be backward-compatible. But it's hard to find a reliable - and nice - way.

